PR #7, opened 2025-01-20
tests: fix Enum initialization

vincentkoppen:
Fixes failing test.

### DIFF
--- a/tests/unit/model/grids/test_grid_search.py
+++ b/tests/unit/model/grids/test_grid_search.py
@@ -13,7 +13,7 @@ from power_grid_model_ds._core.model.enums.nodes import NodeType
 
 def test_grid_get_nearest_substation_node(basic_grid):
     substation_node = basic_grid.get_nearest_substation_node(node_id=103)
-    assert NodeType.SUBSTATION_NODE == NodeType(substation_node.node_type)
+    assert NodeType.SUBSTATION_NODE == NodeType(int(substation_node.node_type))
 
 
 def test_grid_get_nearest_substation_node_no_substation(basic_grid):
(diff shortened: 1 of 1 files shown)


tests/unit/model/grids/test_grid_search.py
```diff
@@ -13,7 +13,7 @@
 
 def test_grid_get_nearest_substation_node(basic_grid):
     substation_node = basic_grid.get_nearest_substation_node(node_id=103)
-    assert NodeType.SUBSTATION_NODE == NodeType(substation_node.node_type)
+    assert NodeType.SUBSTATION_NODE == NodeType(int(substation_node.node_type))
```
Thijss, line 16:
```suggestion
    assert NodeType.SUBSTATION_NODE == NodeType(substation_node.node_type.item())
```
I'd prefer this version. More consistent with the rest of the project.

Thijss, line 16:
```suggestion
    assert NodeType.SUBSTATION_NODE == substation_node.node_type
```
Easier option

vincentkoppen: @mgovers can you check whether this solution also fixes the problem for you? 
If it works I would prefer this one over my own fix.

mgovers: yes it works ✅

jaapschoutenalliander (approved):
Solution is fine by me, I do wonder why this failed only for Martijn, any idea why?

mgovers:
> Solution is fine by me, I do wonder why this failed only for Martijn, any idea why?

maybe Windows and/or Python 3.13?

PR #16, opened 2025-01-23
Feature: add .all_branches property to graph

Thijss:

- Adds new .all_branches property for the Graph model. 
- Useful function when writing tests/debugging since it can be somewhat tricky to obtain the external ids of branches in the graph.


src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -34,6 +34,13 @@ def nr_nodes(self):
     def nr_branches(self):
         """Returns the number of branches in the graph"""
 
+    @property
+    @abstractmethod
+    def all_branches(self) -> list[frozenset[int]]:
+        """Returns all branches in the graph as a list of node pairs (frozensets).
+        Warning: Depending on graph engine, performance could be slow for large graphs
+        """
```
vincentkoppen, line 42: I would like to have list[tuple[int,int]] as the return type.

- I think this is more common within graph libraries (rustworkx also does this)
- A frozenset doesn't specify it's size, while tuples do
- A tuple can be indexed
- It is also quite common that the order doesn't matter when returning for undirected graphs

vincentkoppen (changes requested):

src/power_grid_model_ds/_core/model/graphs/models/base.py
Thijss: Sure, makes sense. Only downside is that with tuples compared to sets it makes it a bit trickier to compare results in some cases

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -34,6 +35,16 @@ def nr_nodes(self):
     def nr_branches(self):
         """Returns the number of branches in the graph"""
 
+    @property
+    def all_branches(self) -> Generator[tuple[int, int], None, None]:
+        """Returns all branches in the graph as a list of node pairs (frozensets).
+        Warning: Depending on graph engine, performance could be slow for large graphs
```
vincentkoppen, line 41: Is this warning still relevant?

vincentkoppen (approved):
Looks good to me! 👍

VERSION
```diff
@@ -1 +1 @@
-1.0
\ No newline at end of file
+1.1
```
Thijss, line 1: Perhaps let's figure out the merging queue before merging

PR #17, opened 2025-01-23
Feature: add tmp_remove_nodes method to graph

Thijss:
Adds a context manager for temporarily removing nodes from the graph.

This can for instance be useful when trying to calculate the shortest path between two nodes without considering certain nodes in the graph.

PR #19, opened 2025-01-24
feat: improve downstream nodes performance with local search

jaapschoutenalliander:
**Performance**: `Improve downstream nodes performance`

> Create downstream nodes function on BaseGraph and RustworkxGraph, restricting the search locally and using the properties of radiality (which is needed anyway to consistently define downstream)

- Add downstream functions
- Rewrite performance tests

### DIFF
--- a/src/power_grid_model_ds/_core/model/graphs/models/base.py
+++ b/src/power_grid_model_ds/_core/model/graphs/models/base.py
@@ -235,7 +235,25 @@ class BaseGraphModel(ABC):
             nodes_to_ignore=self._externals_to_internals(nodes_to_ignore),
             inclusive=inclusive,
         )
+
         return self._internals_to_externals(nodes)
+
+    def get_downstream_nodes(self, node_id: int, stop_node_ids: list[int], inclusive: bool = False) -> list[int]:
+        """Find all nodes connected to the node_id
+        args:
+            node_id: node id to start the search from
+            stop_node_ids: list of node ids to stop the search at
+            inclusive: whether to include the given node id in the result
+        returns:
+            list of node ids sorted by distance, downstream of to the node id
+        """
+        downstream_nodes = self._get_downstream_nodes(
+            node_id=self.external_to_internal(node_id),
+            stop_node_ids=self._externals_to_internals(stop_node_ids),
+            inclusive=inclusive,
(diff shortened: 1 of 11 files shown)


Thijss (commented):
Nice work and I think "sorted by distance" seems preserved. Perhaps we can add/alter a test to verify it? Current tests only seem to check results using `sets`

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -235,8 +235,26 @@ def get_connected(
...
+        returns:
+            list of node ids sorted by distance, downstream of to the node id
+        """
+        downstream_nodes = self._get_downstream_nodes(
+            node_id=self.external_to_internal(node_id),
+            stop_node_ids=self._externals_to_internals(stop_node_ids),
+            inclusive=inclusive,
+        )
+
+        return self._internals_to_externals(downstream_nodes)
```
vincentkoppen, line 256: In what way are the `stop_node_ids` still needed with the `tmp_remove_nodes` context implemented by @Thijss? 
Isn't this equal to: 

```
with graph.tmp_remove_node(stop_node_ids): 
  graph.get_downstream_nodes(node_id = node_id)
```

Thijss: Functionally it should be equal, but since we aren't modifying the graph here it could give significant better performance?
Perhaps could idea to base this PR on #17  and add a performance comparison

jaapschoutenalliander: We could try that, the main performance improve should be in not using get_nearest_substation_node to get_connected on the whole grid. Technically using tmp_remove_nodes can achieve a similar result

jaapschoutenalliander: Issue when we want to do this is that we do not have the information about which substation is the feeder, so I do not think the tmp_remove_nodes will work (without additional complexity)

Thijss:
@jaapschoutenalliander  shall we drop this PR in favor of #21 ?

jaapschoutenalliander:
Agreed, since this solution won't cover all cases

PR #21, opened 2025-01-29
Feat: improve downstream node performance

Thijss:
Improve performance of `grid.get_downstream_nodes`
Also adds a `get_downstream_nodes` graph method

performance difference:
old
Screenshot 2025-02-05 at 15 32 33

new
Screenshot 2025-02-05 at 15 38 49


src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -235,8 +235,41 @@ def get_connected(
...
+            MissingNodeError: if no connected node is found
+            ValueError: if the node_id is in candidate_node_ids
+        """
+        internal_node_id = self.external_to_internal(node_id)
+        internal_candidates = self._externals_to_internals(candidate_node_ids)
+        if internal_node_id in internal_candidates:
+            raise ValueError("node_id cannot be in candidate_node_ids")
+        return self.internal_to_external(self._find_connected(internal_node_id, internal_candidates))
+
+    def get_downstream_nodes(self, node_id: int, stop_node_ids: list[int], inclusive: bool = False) -> list[int]:
```
Thijss, line 258: Still not sure about `stop_node_ids`

Other options:
- start_node_ids
- source_node_ids
- substation_node_ids

all have pros and cons

src/power_grid_model_ds/_core/model/graphs/models/rustworkx.py
```diff
@@ -99,6 +99,13 @@ def _get_connected(self, node_id: int, nodes_to_ignore: list[int], inclusive: bo
 
         return connected_nodes
 
+    def _find_connected(self, node_id: int, candidate_node_ids: list[int]) -> int:
```
jaapschoutenalliander, line 102: I think _find_first_connected would be a more complete description. All candidate_node_ids might be connected, only the first will be returned

src/power_grid_model_ds/_core/model/graphs/models/base.py
jaapschoutenalliander: I do not like source (the term refers to a completely different concept) or substation_node_ids (this function is more general).The keywords should match the expectation of the user, and that makes me like start_node_ids most, since from a user perspective the downstream search starts at these nodes. Stop nodes are more considered from the algorithmic solution point of view.

tests/unit/model/grids/test_grid_search.py
```diff
@@ -25,10 +26,19 @@ def test_grid_get_nearest_substation_node_no_substation(basic_grid):
...
+    basic_grid.make_inactive(basic_grid.link.get(601))
     downstream_nodes = basic_grid.get_downstream_nodes(node_id=102)
-    assert {103, 106} == set(downstream_nodes)
+    assert downstream_nodes[-1] == 104  # Furthest away
+    assert {103, 104, 106} == set(downstream_nodes)
+
+    downstream_nodes = basic_grid.get_downstream_nodes(node_id=102, inclusive=True)
+    assert downstream_nodes[0] == 102
+    assert downstream_nodes[-1] == 104
+    assert {102, 103, 104, 106} == set(downstream_nodes)
```
jaapschoutenalliander, line 41: Can you also add a test for the exception case in the previous PR?

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -235,8 +235,41 @@ def get_connected(
...
+        args:
+            node_id: node id to start the search from
+            stop_node_ids: list of node ids to stop the search at
+            inclusive: whether to include the given node id in the result
+        returns:
+            list of node ids sorted by distance, downstream of to the node id
+        """
+        connected_node = self.find_connected(node_id, stop_node_ids)
+        path, _ = self.get_shortest_path(node_id, connected_node)
+        _, upstream_node, *_ = path  # path is at least 2 elements long or find_connected would have raised an error
```
jaapschoutenalliander, line 269: I like this trick

jaapschoutenalliander (commented):
Good solution, left my opionion on naming

src/power_grid_model_ds/_core/model/graphs/models/base.py
Thijss: I think I also like start_node_ids best

vincentkoppen (approved):

jaapschoutenalliander (approved):

PR #23, opened 2025-01-29
refactor(BaseGraph): improve performance of get_components

vincentkoppen:
**Fixes issue**: Performance of get_components

Use the new `tmp_remove_nodes` within `get_components`, this removes the need of a copy in the rustworkx implementation. 

-


src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -183,7 +183,7 @@ def delete_branch3_array(self, branch_array: Branch3Array, raise_on_fail: bool =
             self.delete_branch_array(branches, raise_on_fail=raise_on_fail)
 
     @contextmanager
-    def tmp_remove_nodes(self, nodes: list[int]) -> Generator:
+    def tmp_remove_nodes(self, nodes: list[int] | NDArray[np.int32]) -> Generator:
```
vincentkoppen, line 186: For now I have left the interface for `get_components` at `NDArray[np.int32]` and added type support here. 
I could also change the argument type of `get_components` if that is what we prefer.

Thijss: For consistency's sake we perhaps let's stick with `list[int]` everywhere for now. Otherwise we could also update all other methods with a `list[int]` input parameter to also support `NDArray[np.int32]`.
Both options are fine with me, but I'd like to apply it consistently at least :)

src/power_grid_model_ds/_core/model/grids/helpers.py
```diff
@@ -45,7 +45,7 @@ def set_feeder_ids(grid: "Grid"):
     601 | 101          | 204
     """
     _reset_feeder_ids(grid)
-    feeder_node_ids = grid.node.filter(node_type=NodeType.SUBSTATION_NODE).id
+    feeder_node_ids = grid.node.filter(node_type=NodeType.SUBSTATION_NODE)["id"]
```
vincentkoppen, line 48: Needed for mypy typing.

src/power_grid_model_ds/_core/model/graphs/models/base.py
vincentkoppen: Done

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -248,9 +247,10 @@ def get_all_paths(self, ext_start_node_id: int, ext_end_node_id: int) -> list[li
 
         return [self._internals_to_externals(path) for path in internal_paths]
 
-    def get_components(self, substation_nodes: NDArray[np.int32]) -> list[list[int]]:
+    def get_components(self, substation_nodes: list[int]) -> list[list[int]]:
```
Thijss, line 250: perhaps we can add the deprecation message for substation_nodes?

something like

```python
if len(substation_nodes):
  # deprecation warning
```

src/power_grid_model_ds/_core/model/graphs/models/rustworkx.py
```diff
@@ -84,11 +84,8 @@ def _get_shortest_path(self, source: int, target: int) -> tuple[list[int], int]:
     def _get_all_paths(self, source: int, target: int) -> list[list[int]]:
         return list(rx.all_simple_paths(self._graph, source, target))
 
-    def _get_components(self, substation_nodes: list[int]) -> list[list[int]]:
-        no_os_graph = self._graph.copy()
```
Thijss, line 88: As @jaapschoutenalliander already pointed out: Surprisingly, there seems to be no performance increase. Verified it on my own laptop. Seem even slightly slower: 0.36 (this PR) vs. 0.33 (main) at 5000 nodes

I still think we can consider merging this because it makes the implementation more readable and more consistent with the other methods. Also, our performance test might not be representative enough. I could try a more representative performance comparison with my internal simulation project at Alliander

vincentkoppen: I verified this myself now as well. Not sure how this can not improve the performance, I guess that the `copy` on a `rustworkx` is implemented very efficiently? 
Tried to tweak the tests a bit as well, but did never get any real difference in performance. 

I would also still opt to merge this, since the implementation is a lot clearer and we can transition to removing these kind of arguments from function with the new `tmp_remove_nodes`.

src/power_grid_model_ds/_core/model/graphs/models/base.py
vincentkoppen: Added deprecation warning. 
image

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -248,9 +248,23 @@ def get_all_paths(self, ext_start_node_id: int, ext_end_node_id: int) -> list[li
...
+        if substation_nodes:
+            warnings.warn(
+                message="""
+get_components: substation_nodes argument is deprecated and will be removed in a future release.
+The functionality is still available by using the `tmp_remove_nodes` context manager.
+
+Example:
+>>> with graph.tmp_remove_nodes(substation_nodes):
+>>>    compontents = graph.get_components()
+""",
```
vincentkoppen, line 262: image

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -248,9 +248,23 @@ def get_all_paths(self, ext_start_node_id: int, ext_end_node_id: int) -> list[li
...
+get_components: substation_nodes argument is deprecated and will be removed in a future release.
+The functionality is still available with the use of the `tmp_remove_nodes` context manager.
+
+Example:
+>>> with graph.tmp_remove_nodes(substation_nodes):
+>>>    components = graph.get_components()
+""",
+                category=DeprecationWarning,
+                stacklevel=2,
+            )
```
vincentkoppen, line 265: image

jaapschoutenalliander (approved):
Clear warning, nicely done

Thijss (approved):

PR #31, opened 2025-02-04
Feat: support adding multiple nodes/branches at once (performance increase)

Thijss:
`rustworkx` supports adding multiple nodes/branches at once. This is much faster than one by one.
I've updated the code to support this.

### benefits
Should increase performance of methods like `Graph.from_arrays` and `Grid.from_cache`

### side-effects
Introduces some breaking changes as some methods on the `BaseGraphModel` have been renamed:
- `add_node` -> `add_node_array`
- `add_branch` -> `add_branch_array`
- `add_branch3` -> `add_branch3_array`  

On the other hand, I don't expect this to break much in existing projects, because in general the `Graph.add_branch` and `Graph.add_node` functions are not directly used (e.g., `Grid.append` and `Grid.add_node` are used instead)

### Performance comparison

Existing code (main):
ExistingCode

New code:
NewCode


src/power_grid_model_ds/_core/model/graphs/container.py
```diff
@@ -83,20 +97,6 @@ def delete_branch3(self, branch: Branch3Array) -> None:
             graph.delete_branch3_array(branch3_array=branch)
             setattr(self, field.name, graph)
 
-    def add_node(self, node: NodeArray) -> None:
```
jaapschoutenalliander, line 86: I would keep the old functions for now with a deprecation warning and a call to the new function

src/power_grid_model_ds/_core/model/graphs/container.py
```diff
@@ -55,18 +55,32 @@ def empty(cls, graph_model: type[BaseGraphModel] = RustworkxGraphModel) -> "Grap
             complete_graph=graph_model(active_only=False),
         )
 
-    def add_branch(self, branch: BranchArray) -> None:
+    def add_node_array(self, node_array: NodeArray) -> None:
+        """Add a node to all graphs"""
+        for field in dataclasses.fields(self):
+            graph = getattr(self, field.name)
+            graph.add_node_array(node_array=node_array, raise_on_fail=False)
+            setattr(self, field.name, graph)
```
jaapschoutenalliander, line 63: Not in scope of this PR, so you may ignore, but I do not think these setattr are neccesary. The reference stays intact

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -93,8 +93,11 @@ def delete_node(self, ext_node_id: int, raise_on_fail: bool = True) -> None:
 
     def add_node_array(self, node_array: NodeArray, raise_on_fail: bool = True) -> None:
         """Add all nodes in the node array to the graph."""
-        for node in node_array:
-            self.add_node(ext_node_id=node.id.item(), raise_on_fail=raise_on_fail)
+        ext_node_ids = node_array["id"].tolist()
+        if raise_on_fail:
+            if existing_ids := set(ext_node_ids).intersection(set(self.external_ids)):
```
jaapschoutenalliander, line 98: The call to external_ids is quite expensive i think, we bypass it on the single add_node by using has_node. Did you consider this, something like `any(self.has_node(x) for x in ext_node_ids)`

Thijss: I think there are pros and cons for both options.
Your suggestion might actually be much slower when adding a large array, since it contains a forloop.

On the other hand, when adding single nodes or small arrays, your suggestion might perform better.

Also I don't think it is that expensive? `.external_ids` are just the keys of the `_external_to_internal` dictionary. But I agree that for single nodes, the has_node method is probably faster.

Perhaps good to do some performance tests with both options



We could compare with some performance

src/power_grid_model_ds/_core/model/graphs/container.py
Thijss: you might be right 👍

src/power_grid_model_ds/_core/model/graphs/models/base.py
Thijss: added your suggestion

jaapschoutenalliander (approved):

PR #33, opened 2025-02-05
Feature: add Grid.from_txt method

Thijss:
- Adds Grid.from_txt method
- Adds support for unordered branches when building a Grid from txt
  - by updating to this: `self.grid.append(new_branch, check_max_id=False)`


src/power_grid_model_ds/_core/model/grids/base.py
```diff
@@ -408,6 +408,17 @@ def _from_pickle(cls, pickle_path: Path):
             raise TypeError(f"{pickle_path.name} is not a valid {cls.__name__} cache.")
         return grid
 
+    @classmethod
+    def from_txt(cls, txt_lines: list[str]):
```
vincentkoppen, line 412: Interface/name wise this might be a bit confusing. 
I would expect the input to be one `str` here, not `list[str]`. 

Possible solutions: 
- Split the string by newlines ourselves in the method
- Change naming: `from_txt_lines`?

vincentkoppen (commented):
Looks good. 
One remark surrounding the interface.

src/power_grid_model_ds/_core/model/grids/base.py
vincentkoppen: Personally I think it would be nice if you could do: 
```
Grid.from_txt("""
S1 2
2 3 
3 5
""")
```

vincentkoppen (approved):

src/power_grid_model_ds/_core/model/grids/base.py
Thijss: I've added suppport for `str`

src/power_grid_model_ds/_core/model/grids/_text_sources.py
```diff
@@ -35,32 +35,33 @@ class TextSource:
...
-                raise ValueError(f"Text line '{text_line}' is invalid. Skipping...") from err
-            comments = comments[0].split(",") if comments else []
+
+            pattern = re.compile(r"^\s*(\S+)\s+(\S+)(?:\s+(\S+))?\s*$")
+            match = pattern.match(text_line)
+            if not match:
+                raise ValueError(f"Text line '{text_line}' is invalid. Skipping...")
+
+            from_node_str, to_node_str, comment = match.groups()
+            comments = comment.split(",") if comment else []
```
vincentkoppen, line 64: Personally I prefer the initial implementation for readability. 
Is this change with the regex needed for anything else?

Issue #27, opened 2025-02-03
[FEATURE] import topology to pgm-ds arrays from dictonary of pgm ComponentType


Bramtimm:
Considering I have a topology from power-grid-model in a dictionary, as from the power-grid-model-example below, I would like to have a simple interface/function to convert this to `pgm-ds` array objects. This would simplify the process of importing `pgm` topologies into `pgm-ds`. 

Given the example below, the function should convert it to a dictionary of `pgm-ds` array objects. Optionally also supporting extension arrays if users have their own array objects.

I am happy to contribute to the implementation if needed.

```
from power_grid_model import ComponentType
from power_grid_model_ds import NodeArray, LineArray, LinkArray, TransformerArray, SourceArray

# PGM-topology
input_data = {
    ComponentType.node: node,
    ComponentType.line: line,
    ComponentType.sym_load: sym_load,
    ComponentType.source: source,
}

topology_pgm_ds = pgm_2_pgm_ds_converter(input_data, ...)

# PGM-DS topology
topology_pgm_ds = {
    "node": NodeArray,
    "line": LineArray,
    "link": LinkArray,
    "transformer": TransformerArray,
    "source":  SourceArray,
}
```

jaapschoutenalliander:
I agree with that it is a good feature to have. It makes integrating with the pgm-io conversion output easier as well. The design idea for adding it to the exesiting PGMInterface class is a good choice

PR #39, opened 2025-02-11
Feat: Create PGM-DS Grid object from PGM dict

MvLieshout:
Signed-off-by: Marnix van Lieshout <marnix.van.lieshout@alliander.com>

**Fixes issue**: #27

- Added method to create a PGM-DS Grid object from a PGM dict.
- Test
- Example in notebook


src/power_grid_model_ds/_core/load_flow.py
```diff
@@ -63,6 +64,23 @@ def create_input_from_grid(self):
...
+        array_mapping: Dict[str, Type[FancyArray]],
+    ) -> Grid:
+        """
+        Create Grid object from PowerGridModel input.
+        Note that for some arrays, not all fields are available in the PowerGridModel input. 
+        In this case, the default values are used.
+
+        Returns a Grid object with the arrays filled with the PowerGridModel input.
+        """
+        for pgm_name, pgm_ds_array_class in array_mapping.items():
```
jaapschoutenalliander, line 78: If the branch arrays are before the node array this will fail at the moment, because branches can only be added between nodes present in the grid. I would suggest using the PGM_ARRAYS object to fix the order

src/power_grid_model_ds/_core/load_flow.py
```diff
@@ -63,6 +64,23 @@ def create_input_from_grid(self):
...
+            if pgm_name in self.input_data:
+                pgm_ds_array = pgm_ds_array_class(self.input_data[pgm_name])
```
jaapschoutenalliander, line 80: In continuation of the above comment, here you can then use something like
```
pgm_ds_array_class = getattr(grid, pgm_name).__class__
```

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +453,93 @@ def test_batch_run():
...
+    }
+
+    core_interface = PowerGridModelInterface(input_data=input_data)
+    output = core_interface.create_grid_from_input_data(
+        array_mapping=array_mapping,
+    )
+
+    assert isinstance(output, Grid)
+    assert len(output.node) == len(grid.node)
+    # NOTE: The grid object is not equal to the output object for some default values.
```
jaapschoutenalliander, line 481: Also consider asserting these values are consistent with the default

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +453,93 @@ def test_batch_run():
 
     # Results have been calculated for all 10 scenarios
     assert 10 == len(output["line"])
+
+
+def test_create_grid_from_input_data():
```
jaapschoutenalliander, line 458: I would also like a test on what happens when an extended array does not contain a default value. This would result in an error, which is fine in my opinion

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +453,93 @@ def test_batch_run():
...
+
+    core_interface = PowerGridModelInterface(input_data=input_data)
+    output = core_interface.create_grid_from_input_data(
+        array_mapping=array_mapping,
+    )
+
+    assert isinstance(output, Grid)
+    assert len(output.node) == len(grid.node)
+    # NOTE: The grid object is not equal to the output object for some default values.
+    assert np.array_equal(grid.node[["id", "u_rated"]], output.node[["id", "u_rated"]])
```
jaapschoutenalliander, line 482: Maybe take the column values from the input_data arrays dtypes intersected with grid array dtype? To make it clear these column should always match. You could use the `_match_dtypes` on the `PowerGridModelInterface` class

jaapschoutenalliander (changes requested):
Thank you for setting up the PR, it is a good set up in my opinion.

On your questions,
- For me making the grid object optional is fine
- Using the default values is the only option I can see for creating a full array from partial information in the pgm-input

My main concern for this solution is the user provided array_mapping, I left a suggestion on that

MvLieshout:
Thanks for the feedback @jaapschoutenalliander, I will work on making some changes to the code over de next days.

tests/integration/loadflow/test_power_grid_model.py
MvLieshout: For my use-case this would actually be a problem I think. In our ideal case, this would also work for extended arrays. I wil try to find a way to do this.

jaapschoutenalliander: Yes indeed

jaapschoutenalliander: Also, I do think there could be a future extension where you provide extra_info in the format of https://github.com/PowerGridModel/power-grid-model-io/blob/cb1f8211c4dba02309165c8066837d815c00b4f5/src/power_grid_model_io/data_types/_data_types.py#L10 to fill these columns with other values than the default

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +455,125 @@ def test_batch_run():
...
+        ),
+    )
+
+    assert type(output.source) is SourceArray
+    assert np.array_equal(
+        output.source.data,
+        np.array([(8, 7, 1, 1.0)], dtype=[("id", "<i4"), ("node", "<i4"), ("status", "i1"), ("u_ref", "<f8")]),
+    )
+
+class CustomNodeArrayDefault(NodeArray):
```
MvLieshout, line 532: Not sure where these should be placed.

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +455,125 @@ def test_batch_run():
...
     # Results have been calculated for all 10 scenarios
     assert 10 == len(output["line"])
+
+
+def test_create_grid_from_input_data(input_data_pgm):
+    core_interface = PowerGridModelInterface(input_data=input_data_pgm)
+    output = core_interface.create_grid_from_input_data()
+
+    assert isinstance(output, Grid)
+    assert type(output.node) is NodeArray
```
jaapschoutenalliander, line 465:
```suggestion
    assert isinstance(output.node, NodeArray)
```

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +455,125 @@ def test_batch_run():
...
+                ("id", "<i4"),
+                ("u_rated", "<f8"),
+                ("node_type", "i1"),
+                ("feeder_branch_id", "<i4"),
+                ("feeder_node_id", "<i4"),
+            ],
+        ),
+    )
+
+    assert type(output.line) is LineArray
```
jaapschoutenalliander, line 484:
```suggestion
    assert isinstance(output.line, LineArray)
```

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +455,125 @@ def test_batch_run():
...
+                ("r1", "<f8"),
+                ("x1", "<f8"),
+                ("c1", "<f8"),
+                ("tan1", "<f8"),
+                ("i_n", "<f8"),
+            ],
+        ),
+    )
+
+    assert type(output.sym_load) is SymLoadArray
```
jaapschoutenalliander, line 510:
```suggestion
    assert isinstance(output.sym_load, SymLoadArray)
```

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +455,125 @@ def test_batch_run():
...
+                ("node", "<i4"),
+                ("status", "i1"),
+                ("type", "i1"),
+                ("p_specified", "<f8"),
+                ("q_specified", "<f8"),
+            ],
+        ),
+    )
+
+    assert type(output.source) is SourceArray
```
jaapschoutenalliander, line 526:
```suggestion
    assert isinstance(output.source, SourceArray)
```

tests/integration/loadflow/test_power_grid_model.py
jaapschoutenalliander: I think this if fine, we often use custom definitions with specific tests

src/power_grid_model_ds/_core/load_flow.py
```diff
@@ -63,6 +63,23 @@ def create_input_from_grid(self):
...
+        In this case, the default values are used.
+
+        Returns a Grid object with the arrays filled with the PowerGridModel input.
+        """
+        for pgm_name in PGM_ARRAYS:
+            if pgm_name in self.input_data:
+                pgm_ds_array_class = getattr(self.grid, pgm_name).__class__
+                pgm_ds_array = pgm_ds_array_class(self.input_data[pgm_name])
+                self.grid.append(pgm_ds_array, check_max_id=False)
+        return self.grid
```
jaapschoutenalliander, line 81:
```suggestion
    def create_grid_from_input_data(self, check_ids: bool = True) -> Grid:
        """
        Create Grid object from PowerGridModel input.
        Note that for some arrays, not all fields are available in the PowerGridModel input.
        In this case, the default values are used.

        Args:
            check_ids: if True, check if the ids are unique

        Returns a Grid object with the arrays filled with the PowerGridModel input.
        """
        for pgm_name in PGM_ARRAYS:
            if pgm_name in self.input_data:
                pgm_ds_array_class = getattr(self.grid, pgm_name).__class__
                pgm_ds_array = pgm_ds_array_class(self.input_data[pgm_name])
                self.grid.append(pgm_ds_array, check_max_id=False)
        if check_ids:
            self.grid.check_ids()
        return self.grid
```

I suggest adding a check on ids option, since the grid object is not necessarily empty. This I think makes it safe enough

jaapschoutenalliander (commented):
Looks good, I have made some suggestions after which I think it is complete, mainly in test syntax

src/power_grid_model_ds/_core/load_flow.py
Thijss: Just to note: this way the grid will already have duplicated ids by the time check_ids() raises an error.

I still think it is fine to go with this solution

tests/integration/loadflow/test_power_grid_model.py
```diff
@@ -453,3 +455,125 @@ def test_batch_run():
 
     # Results have been calculated for all 10 scenarios
     assert 10 == len(output["line"])
+
+
+def test_create_grid_from_input_data(input_data_pgm):
```
Thijss, line 460: lets group these tests under a class like `class TestCreateGridFromInputData:`

For an example check here:
https://github.com/PowerGridModel/power-grid-model-ds/blob/938307d4e46f40db5ac152acb85df6e317d4e5db/tests/unit/model/graphs/test_graph_model.py#L18

src/power_grid_model_ds/_core/load_flow.py
jaapschoutenalliander: It is correct that in this setup the check will raise an error while having created an 'invalid grid', I think this is fine since it is not an expected path which needs recovering from.

On the alternative, that will block more than just non-unique ids since arrays are added one-by-one and no order in ids is expected. So if the node id has 2 and the line id has 1 that would raise an error while the ids are fine.

jaapschoutenalliander (approved):
Thank you for adding the feature, we appreciate the effort!

Before merging I have one last request, to update the VERSION file to 1.2 https://github.com/PowerGridModel/power-grid-model-ds/blob/main/VERSION (also for your own ease of implementation)

PR #41, opened 2025-02-26
chore: add typing to tests

jaapschoutenalliander:
Adds typing to tests for ease of development


src/power_grid_model_ds/_core/model/grids/base.py
```diff
@@ -51,7 +52,7 @@
 
 
 @dataclass
-class Grid(FancyArrayContainer):
+class Grid(FancyArrayContainer, MinimalGridArrays):
```
vincentkoppen, line 55: Do we want to keep the `MinimalGridArrays` or switch to a solution with `if TYPE_CHECKING: ...` to work around the circular dependency?

vincentkoppen (approved):
Looks good to me, one question that is not directly related with this PR

PR #43, opened 2025-03-11
tempTrary PR to show performance difference between two implementations

Thijss:
before suggestion

Screenshot 2025-03-11 at 18 01 25

after suggestion

Screenshot 2025-03-11 at 18 02 03

### DIFF
--- a/src/power_grid_model_ds/_core/model/graphs/models/base.py
+++ b/src/power_grid_model_ds/_core/model/graphs/models/base.py
@@ -93,11 +93,10 @@ class BaseGraphModel(ABC):
 
     def add_node_array(self, node_array: NodeArray, raise_on_fail: bool = True) -> None:
         """Add all nodes in the node array to the graph."""
-        ext_node_ids = node_array["id"].tolist()
-        if raise_on_fail:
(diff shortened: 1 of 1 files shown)


src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -93,11 +93,10 @@ def delete_node(self, ext_node_id: int, raise_on_fail: bool = True) -> None:
 
     def add_node_array(self, node_array: NodeArray, raise_on_fail: bool = True) -> None:
         """Add all nodes in the node array to the graph."""
-        ext_node_ids = node_array["id"].tolist()
-        if raise_on_fail:
-            if existing_ids := set(ext_node_ids).intersection(set(self.external_ids)):
-                raise GraphError(f"{len(existing_ids)} external node ids already exist!")
-        self._add_nodes(ext_node_ids)
+        if raise_on_fail and any(self.has_node(x) for x in node_array["id"]):
+            existing_ids = [x for x in node_array["id"] if self.has_node(x)]
```
jaapschoutenalliander, line 97: This line could be skipped and just given a general warning, for what it's worth

jaapschoutenalliander (commented):
So no huge difference, but some difference...

PR #46, opened 2025-03-20
fix minor doc typos

mgovers:
Fix a couple small typos during full search through docs.

### DIFF
--- a/docs/examples/utils/grid_from_txt_examples.ipynb
+++ b/docs/examples/utils/grid_from_txt_examples.ipynb
@@ -20,7 +20,7 @@
     "If you follow a specific syntax, you can use the input from the\n",
     "graph editor to transfer your drawing to pgm-ds:\n",
     "\n",
-    "- A _source node_ should be prefixed with `S`\n",
+    "- A _substation node_ should be prefixed with `S`\n",
     "  - e.g: `S1`, `S2`\n",
     "- A _line_ is defined as `<from_node> <to_node>`\n",
     "  - From nodes can be both regular nodes and source nodes\n",
--- a/docs/examples/utils/grid_from_txt_examples.ipynb
+++ b/docs/examples/utils/grid_from_txt_examples.ipynb
@@ -61,7 +61,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 9,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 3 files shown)


docs/examples/utils/grid_from_txt_examples.ipynb
```diff
@@ -84,29 +84,27 @@
...
-    "        \"8 9\",\n",
-    "    ]\n",
+    "    \"S1 2\",\n",
+    "    \"S1 3 open\",\n",
+    "    \"2 7\",\n",
+    "    \"3 5\",\n",
+    "    \"3 6 transformer\",\n",
+    "    \"5 7\",\n",
+    "    \"7 8\",\n",
+    "    \"8 9\",\n",
```
mgovers, line 101: another problem I found is that sources and loads are not created - only transformers and lines, even though the docs say that "source nodes" exist and are prefixed `S`

Thijss: The docs are outdated. If I recall correctly, I removed that functionality a while back in a refactor PR of the TxtGrid logic, due to I believe a suggestion by @vincentkoppen.

As for the reason, I'm not too sure anymore to be honest. 
In my team's project, we supercharge the method and add the loads/sources ourselves as well 😅.

I'd be fine with re-adding the functionality.
Currently the only thing the 'S' alters, is updating the node's `.node_type` to `NodeType.SUBSTATION_NODE`

Thijss (approved):

docs/examples/utils/grid_from_txt_examples.ipynb
mgovers: I do like the functionality to automatically add sources and loads, but let's make sure to first get the docs up-to-date again. Do you want me to do that in this PR or in a separate one?

mgovers: updated the docs to reflect prefix `S` means `substation node`

docs/model_interface.ipynb
mgovers, line 1: one more thing i've found is that the documentation doesn't mention loads a lot.

E.g. the first time "load" is mentioned in the random grid generator example is only in the print statements of the first cell: `print(f"Number of loads: {len(grid.sym_load)}")` and the `quickstart` doesn't even mention them at all.

Since it affects more than just simple typos, I didn't want to do it in this PR, but it also feels too small to create a separate issue. What do you think?

Thijss: I think we can add documentation on loads in a next PR. Could be a nice first issue for someone to pick up

Thijss: I've created an issue https://github.com/PowerGridModel/power-grid-model-ds/issues/48

PR #53, opened 2025-04-18
Feat/prevent auto major minor release

Thijss:
Last minor release (1.2) it was difficult to group multiple PRs into one minor release using a merge queue. Reason: the queue only works if you merge PRs quick enough after another, which can be difficult when dealing with merge conflicts.

Therefore, I've altered the release workflow so that major and minor releases can only be triggered manually.

Should work! :sweat_smile:


jaapschoutenalliander (approved):
How would you then forse a manually triggered major/minor release? I think it is good to document this, and lets try it to see if it works now

Thijss:
> How would you then forse a manually triggered major/minor release? I think it is good to document this, and lets try it to see if it works now

Documentation is a good suggestion.
You can do a major/minor release by starting the [workflow](https://github.com/PowerGridModel/power-grid-model-ds/actions/workflows/build-test-and-sonar.yml) from the `main` branch

PR #54, opened 2025-04-18
Fix: re-order publish steps so artifact isn't deleted by checkout step

Thijss:
It seems that `actions/checkout@v4` resets the working directory and thus deletes `wheelhouse/`. I expect that re-ordering the steps will fix the issue.


Thijss:
Fun fact: deduced this myself, then asked CoPilot to assess the issue and it came up with the same solution. Guess I'm retiring soon.. ;)

jaapschoutenalliander (approved):

PR #60, opened 2025-05-08
Fix: prevent type change when using tmp_remove_nodes

Thijss:
Found was an issue where the graph.external_ids would change type if the input to graph.tmp_remove_nodes is an ndarray[int[ instead of a list[int]

While a type-checker should prevent you from doing this. It is not enforced at runtime.
This PR prevents the type change by explicitly casting the ids to integers before re-adding them to the graph.

Example:
```python
    with graph.tmp_remove_nodes(np.array([1, 2])): 
        pass
# would previously result in graph.external_ids to contain np.int values instead of int values
```

### DIFF
--- a/src/power_grid_model_ds/_core/model/graphs/models/base.py
+++ b/src/power_grid_model_ds/_core/model/graphs/models/base.py
@@ -208,7 +208,7 @@ class BaseGraphModel(ABC):
         yield
 
         for node in nodes:
-            self.add_node(node)
+            self.add_node(int(node))  # convert to int to avoid type issues when input is e.g. a numpy array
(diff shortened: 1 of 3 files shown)


Thijss:
Open question: should we change the type signature of `def tmp_remove_nodes(self, nodes: list[int]) -> Generator:` to also allow `ndarray[int]`?
I opted for no so far since we previously settled on keeping everything `list[int]` within the Graph class

jaapschoutenalliander (approved):
Well found. On the typing, I think it's cleaner to keep that as it is and just see support for other iterables as a bonus

PR #68, opened 2025-05-30
feat: setup three winding trafo visualisation

jaapschoutenalliander:
Draft for visualisation of three winding transformers
image

Choices to discus:
- The three winding transformer is visualized by 3-connections between the nodes. Could also be done using a helper node. However, the current choice is based on the internal graph modelling choice
- Did not add a new visualization style for 3-winding-transformer but visualized as transformer (blue lines)


src/power_grid_model_ds/_core/visualizer/parsers.py
```diff
@@ -32,9 +32,48 @@ def parse_branches(grid: Grid) -> list[dict[str, Any]]:
...
+                    "source": str(branch_.from_node.item()),
+                    "target": str(branch_.to_node.item()),
+                    "group": group,
+                }
+            )
+            parsed_branches.append(cyto_elements)
+    return parsed_branches
+
+
+def _branch3_to_branches(branch3: Branch3Array) -> BranchArray:
```
jaapschoutenalliander, line 59: Duplicate code, should be centralized

PR #76, opened 2025-07-16
Feat: add from_extended methods

Thijss:
Adds

```python
base_node_array = LineArray.from_extended(my_extended_line_array)
```
And
```python
base_grid = Grid.from_extended(my_extended_grid)
```

### DIFF
--- a/src/power_grid_model_ds/_core/model/arrays/base/array.py
+++ b/src/power_grid_model_ds/_core/model/arrays/base/array.py
@@ -323,3 +323,13 @@ class FancyArray(ABC):
         if pandas is None:
             raise ImportError("pandas is not installed")
         return pandas.DataFrame(self._data)
+
+    @classmethod
+    def from_extended(cls: Type[Self], extended: Self) -> Self:
+        """Create an instance from an extended array."""
+        if not isinstance(extended, cls):
+            raise TypeError(f"Extended array must be of type {cls.__name__}, got {type(extended).__name__}")
+        new_array = cls.zeros(len(extended), empty_id=False)
+        for column in cls.get_dtype().names:
+            new_array[column] = extended[column]
+        return new_array
--- a/src/power_grid_model_ds/_core/model/arrays/base/array.py
+++ b/src/power_grid_model_ds/_core/model/arrays/base/array.py
@@ -329,7 +329,5 @@ class FancyArray(ABC):
         """Create an instance from an extended array."""
         if not isinstance(extended, cls):
             raise TypeError(f"Extended array must be of type {cls.__name__}, got {type(extended).__name__}")
-        new_array = cls.zeros(len(extended), empty_id=False)
-        for column in cls.get_dtype().names:
-            new_array[column] = extended[column]
-        return new_array
+        dtype = cls.get_dtype()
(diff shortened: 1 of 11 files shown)


src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -323,3 +323,13 @@ def as_df(self: Self):
...
         return pandas.DataFrame(self._data)
+
+    @classmethod
+    def from_extended(cls: Type[Self], extended: Self) -> Self:
+        """Create an instance from an extended array."""
+        if not isinstance(extended, cls):
+            raise TypeError(f"Extended array must be of type {cls.__name__}, got {type(extended).__name__}")
+        new_array = cls.zeros(len(extended), empty_id=False)
+        for column in cls.get_dtype().names:
+            new_array[column] = extended[column]
```
vincentkoppen, line 334: I prefer the use of `empty` over `zeros` in general for these kind of construction. Now we set zero values that could actually have influence in a calculation possibly (or set strange references)
We probably overwrite all values here, so shouldn't cause any actual problems? 

Alternatively we have this code on our side to do the same (tweaked to this use case): 
```
dtype = cls.get_dtype()
return cls(data=np.array(extended[list(dtype.names)], dtype=dtype))
```

src/power_grid_model_ds/_core/model/grids/base.py
```diff
@@ -440,6 +440,23 @@ def set_feeder_ids(self):
         set_is_feeder(grid=self)
         set_feeder_ids(grid=self)
 
+    @classmethod
+    def from_extended(cls, extended: "Grid", load_graphs: bool = True) -> "Grid":
```
vincentkoppen, line 444: Why would we want to allow `load_graphs = False`?

src/power_grid_model_ds/_core/model/grids/base.py
```diff
@@ -440,6 +440,23 @@ def set_feeder_ids(self):
...
+        for field in dataclasses.fields(cls):
+            if issubclass(field.type, FancyArray):
+                extended_array = getattr(extended, field.name)
+                new_array = field.type.from_extended(extended_array)
+                setattr(new_grid, field.name, new_array)
+
+        new_grid._id_counter = new_grid.max_id
+
+        if load_graphs:
+            new_grid.graphs = GraphContainer.from_arrays(new_grid)
```
vincentkoppen, line 457: Would prefer to use `append` (with `check_max_id=False`)
Then you also always update the graph.

tests/unit/model/arrays/test_array.py
```diff
@@ -289,3 +289,10 @@ def test_overflow_value():
     with pytest.raises(OverflowError):
         transformer.tap_min = -167
     assert transformer.tap_min == -128
+
+
+def test_from_extended_array():
+    extended_array = ExtendedLineArray.zeros(2)
```
vincentkoppen, line 295: Would prefer a test with some actual data in the array. Now the copying over of values isn't tested

vincentkoppen (commented):
Some remarks, overal good functionality to support! 
Do we want these to the docs/example notebooks?

jaapschoutenalliander (approved):

PR #77, opened 2025-07-16
Add Grid.from_pgm_data

Thijss:
Adds a helper method that simplifies creating a grid from PGM input data

also made  `array_equal_with_nan`public and moved to `misc` module


jaapschoutenalliander:
The PowerGridModelInterface already supports this conversion, see https://power-grid-model-ds.readthedocs.io/en/stable/examples/pgm/basic_pgm_examples.html#create-grid-from-input-data-example

So could we reuse these functions or do they suffice in the current implementation?

PR #83, opened 2025-07-30
Reorganize visualizer menu and add show-arrows option

### LINKED ISSUE #63: [IMPROVEMENT] Remove arrows in visualzier or make them optional
Based on feedback from the workshop session at the PGM meetup, I think it is a good idea to remove the arrows in the visualized grid or (at least) make them optional.

Thijss:
https://github.com/user-attachments/assets/9e60fb09-ea8c-44de-88e4-4053740a8453

Note: when you hide arrows and switch back and forth to another menu (e.g. config > legend > config) The arrows will be visible again. Couldn't figure out an easy way to prevent that.
I have some ideas, but they'll require more work. Ah well, good enough for now..

fixes #63


src/power_grid_model_ds/_core/visualizer/layout/cytoscape_styling.py
```diff
@@ -4,8 +4,13 @@
...
 
+from typing import Any
+
 from power_grid_model_ds._core.visualizer.layout.colors import CYTO_COLORS
 
+STYLESHEET = list[dict[str, Any]]
+
+
 NODE_SIZE = 100
 BRANCH_WIDTH = 10
```
vincentkoppen, line 15: Should we set STYLESHEET in a new `typing.py` file? 
I found it a bit odd that this is a type and not a constant from it's name and import in fdiles as well. 

If the file name indicates it is a type, it is much more clear I think.

src/power_grid_model_ds/_core/visualizer/layout/header.py
```diff
@@ -3,29 +3,59 @@
...
-    "border-left": "1px solid white",
-    "border-right": "1px solid white",
-    "border-radius": 0,
-}
-
+SHOW_ARROWS_CHECKBOX = dbc.Checkbox(
+    id="show-arrows",
+    label="Show arrows",
+    value=True,
+    className="me-2",
```
vincentkoppen, line 16: Can this me a bit more descriptive?

vincentkoppen (commented):
Looks good, nice improvement. 
Some small nitpicks, see whether you agree. Otherwise most of them can be ignored.

src/power_grid_model_ds/_core/visualizer/layout/header.py
Thijss: removed

src/power_grid_model_ds/_core/visualizer/layout/cytoscape_styling.py
Thijss: done

vincentkoppen (approved):
Looks good to me! 
Sonarcloud has some issues, mainly duplicate code in the class names.

Issue #45, opened 2025-03-20
[FEATURE] Add GenericBranch support


mgovers:
The GenericBranch (supported by the `power-grid-model`, cfr. https://power-grid-model.readthedocs.io/en/stable/user_manual/components.html#generic-branch) is not supported according to the docs (https://power-grid-model-ds.readthedocs.io/en/stable/api_reference/power_grid_model_ds.html#power_grid_model_ds.arrays.BranchArray). It would be great to enable support for it, as it it has an interface that is closer to other data formats like CIM/CGMES.

jaapschoutenalliander:
I agree this would be a good addition and complete the supported branch types. It should be easy to do, needed is
- Adding a definition here https://github.com/PowerGridModel/power-grid-model-ds/blob/main/src/power_grid_model_ds/_core/model/dtypes/branches.py
- Defining the corresponding array https://github.com/PowerGridModel/power-grid-model-ds/blob/main/src/power_grid_model_ds/_core/model/arrays/pgm_arrays.py
- Adding the array to the grid https://github.com/PowerGridModel/power-grid-model-ds/blob/main/src/power_grid_model_ds/_core/model/grids/base.py

Thijss:
extra ToDo once #47 is merged:
- Add support for GenericBranch in new visualizer

mgovers:
As a follow-up to the LF Energy meet-up, I would like to bump this feature request once more. In addition to GenericBranch support, I am also thinking about:

* [ ] Generic Branch (this issue)
* [ ] Local and Global current sensors (a.o. for PMU support)
* [ ] Asymmetric Line

Should I create separate GitHub issues for that?

mgovers:
> I think it is fine to keep them together, I opened a PR which adds them. Do you consider all the array attributes neccesary in the base model or are the attributes we could leave out?

* for generic branch
  * I'd consider `sn` (rated power) to be as optional as it is for regular transformers. Since it's present there, please keep it.
* for asym line
  * i'd consider only `r_aa, ..., r_cc`, `x_aa, ..., x_cc`, `c0` and `c1` to be required.
  * the rest is quite a complex combination of supported features (see https://power-grid-model.readthedocs.io/en/stable/user_manual/components.html#id11)
  * note that `c0` and `c1` are only optional if and only if the other `c_aa, ..., c_cc` are not provided.
[...]

PR #94, opened 2025-10-01
feat: add extra components from pgm

jaapschoutenalliander:
Implements https://github.com/PowerGridModel/power-grid-model-ds/issues/45


src/power_grid_model_ds/_core/model/dtypes/branches.py
```diff
@@ -43,6 +43,28 @@ class Line(Branch):
...
+    r1: NDArray[np.float64]  # positive-sequence resistance
+    x1: NDArray[np.float64]  # positive-sequence reactance
+    g1: NDArray[np.float64]  # positive-sequence conductance
+    b1: NDArray[np.float64]  # positive-sequence susceptance
+    k: NDArray[np.float64]  # off-nominal ratio
+    theta: NDArray[np.float64]  # angle shift (radian)
+    sn: NDArray[np.float64]  # rated power
+
+    _defaults = {
+        "k": 1.0,
```
Thijss, line 62: why add the defaults?

jaapschoutenalliander: These are defaults on the pgm side as well. I adopted those, fine with leaving them out as well

Thijss: Ah I see. https://power-grid-model.readthedocs.io/en/stable/user_manual/components.html#id9

In that case fine with me to keep them

src/power_grid_model_ds/_core/visualizer/layout/colors.py
```diff
@@ -2,15 +2,20 @@
...
 
+
 YELLOW = "#facc37"
 CYTO_COLORS = {
     "line": YELLOW,
     "link": "#008000",
     "transformer": "#4290f5",
     "node": YELLOW,
+    "generic_branch": "#dddddd",
+    "asym_line": "#2de2ca",
```
nitbharambe: I'm curious, the same reasoning mentioned for `three_winding_transformer` at https://github.com/PowerGridModel/power-grid-model-ds/pull/68#pullrequestreview-3029927664 would be applicable here as well right? 
I imagine PGM team can add more Branch or Branch3 types in future if requested by the community.

(No requirement from me about the new styles. Plus I imagine we don't have to support keeping same styles in future.)

jaapschoutenalliander: The layout has somewhat changed to have more space but it stays a balance. I think for now having all transformers the same color is fine since they are recognisable by shape

PR #113, opened 2025-11-12
PR on #94: add AsymPowerSensorArray and remove optional columns from GenericBranch

Thijss:
Add `AsymPowerSensorArray` and remove defaults from `GenericBranch`


tests/unit/model/arrays/test_build.py
```diff
@@ -232,27 +232,13 @@ def test_initialization_from_args_with_extra_columns():
...
-        "power_sigma",
-        "p_measured",
-        "p_sigma",
-        "q_measured",
-        "q_sigma",
-    ]
-    for field in sym_pow_fields:
-        assert field in pow_sens.dtype.fields
-
+def test_asymmetric_sensor_array():
```
Thijss: the first part of the original test did't really test anything useful

PR #115, opened 2025-11-12
Use getitem everywhere

Thijss:
Replace `array.field_name` with `array["field_name"]`

ToDo:
- [ ] run performance tests and compare with main to be sure

### DIFF
--- a/src/power_grid_model_ds/_core/model/grids/base.py
+++ b/src/power_grid_model_ds/_core/model/grids/base.py
@@ -158,9 +158,9 @@ class Grid(FancyArrayContainer):
             raise ValueError("No branch_ids provided.")
         for branch_array in self.branch_arrays:
             array = branch_array.filter(branch_ids)
-            if 0 < len(array) != len(branch_ids):
+            if array.size > 0 and array.size != len(branch_ids):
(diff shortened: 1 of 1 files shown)


src/power_grid_model_ds/_core/data_source/generator/arrays/base.py
```diff
@@ -12,7 +12,7 @@
 class BaseGenerator:
     """Base class to build a generator for grid elements"""
 
-    def __init__(self, grid: Grid, seed: int) -> None:
+    def __init__(self, grid: Grid, seed: int | None) -> None:
```
Thijss: piggyback: This is a valid option and the current default in the RadialGridGenerator

src/power_grid_model_ds/_core/fancypy.py
```diff
@@ -26,7 +26,15 @@ def concatenate(fancy_array: T, *other_arrays: Union[T, np.ndarray]) -> T:
     return fancy_array.__class__(data=concatenated)
 
 
-def unique(array: T, **kwargs):
+@overload
```
Thijss: piggyback: add overload

PR #116, opened 2025-11-12
Fix broken performance test

Thijss:
Probably broken by #92


tests/performance/grid_performance_tests.py
```diff
@@ -57,6 +57,6 @@ def perf_test_get_downstream_nodes_performance():
 
 
 if __name__ == "__main__":
-    perf_test_get_downstream_nodes_performance()
+    # perf_test_get_downstream_nodes_performance()
```
Thijss:
```suggestion
    perf_test_get_downstream_nodes_performance()
```

PR #122, opened 2025-12-05
Refactor: internal grid modules

Thijss:
Move the logic of Grid methods to separate submodules.
In my opinion, the Grid class was becoming quite large. With this PR the code is better organized into separate modules.

Additionally, I've deprecated add_node & add_branch in favor of grid.append
Piggyback: moved set_is_feeder into set_feeder_ids (it wasn't used anywhere else)
Piggyback: switch order of poe all tasks, so formatting is done before linting


src/power_grid_model_ds/_core/model/grids/_feeders.py
```diff
@@ -44,6 +44,7 @@ def set_feeder_ids(grid: "Grid"):
     301 | 101          | 201
     601 | 101          | 204
     """
+    _set_is_feeder(grid=grid)
```
Thijss: extra: moved this inside set_feeder_ids

PR #130, opened 2026-01-12
Python support: add support for 3.14

mgovers:
PGM has been supporting python 3.14 since its release, but PGM-DS has had a hard version requirement rejecting python 3.14 and newer. This PR add support for python 3.14

Since PGM-DS in its current form does not have any code that forces it to restrict its maximum python version, we remove the maximum version restriction.

Since Python 3.11 support will be dropped soon after, a specific patch version shall be released in-between these two versions to allow users to migrate.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ classifiers = [
   "Operating System :: MacOS",
   "Topic :: Scientific/Engineering :: Physics",
 ]
-requires-python = ">=3.11,<3.15"
+requires-python = ">=3.11"
 dependencies = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers = [
   "Operating System :: MacOS",
   "Topic :: Scientific/Engineering :: Physics",
 ]
-requires-python = ">=3.11,<3.14"
+requires-python = ">=3.11,<3.15"
```
Thijss:
```suggestion
requires-python = ">=3.11"
```

Issue #118, opened 2025-11-19
[FEATURE] Merge grids


Thijss:
It would be nice if we could merge multiple grids via e.g. the following syntax:

```python
grid_1_and_2 = grid_1.merge(grid_2)
```
The challenge here is to avoid id conflicts in the resulting grid.
The proposal is to regenerate all ids to avoid conflicts.

Note 1:  regenerating all ids also means that all references to ids (i.e. node ids) need to be updated too.

Note 2: Teams that require a UUID on assets across all grids, need to implement this themselves by [extending the Grid](https://power-grid-model-ds.readthedocs.io/en/stable/examples/model/grid_extensions_examples.html#grid-extension-examples)

PR #136, opened 2026-01-13
feat: support merge of two grids

wwadman:
**Fixes issue**: `#118`

Adding function Grid.merge(self, other_grid) that merges another grid into this grid.

1. I went for the hardcoded approach as explained in https://github.com/PowerGridModel/power-grid-model-ds/issues/118#issuecomment-3552540029
2. Ideally we should make this more general, for example by adding an indicator ID_COLUMN whether a column contains ids or not.

~Search for `Question for reviewer` in the diff~

### Checks
I added some tests that in my view cover most of the workings.

### DIFF
--- a/src/power_grid_model_ds/_core/model/grids/_helpers.py
+++ b/src/power_grid_model_ds/_core/model/grids/_helpers.py
@@ -63,7 +63,7 @@ def create_empty_grid(grid_class: Type[G], graph_model: type[BaseGraphModel] = R
 def merge_grids(grid: G, other_grid: G, mode: Literal["recalculate_ids", "keep_ids"]) -> None:
     """See Grid.merge()"""
 
-    if type(grid) != type(other_grid):
+    if type(grid) is not type(other_grid):
         raise TypeError("Both grids should be of the same class (to ensure they have the same arrays)")
 
     other_grid_all_arrays = list(other_grid.all_arrays())


src/power_grid_model_ds/_core/model/grids/_helpers.py
```diff
@@ -42,3 +57,50 @@ def create_empty_grid(grid_class: Type[G], graph_model: type[BaseGraphModel] = R
     empty_fields = grid_class._get_empty_fields()  # noqa # pylint: disable=protected-access
     empty_fields["graphs"] = GraphContainer.empty(graph_model=graph_model)
     return grid_class(**empty_fields)
+
+
+def merge_grids(grid: G, other_grid: G) -> G:
+    """See Grid.merge()"""
+
+    # Question for reviewer: below 2 deep copies keep the 2 original grids intact, is there a more elegant way?
+    new_grid = copy.deepcopy(grid)
```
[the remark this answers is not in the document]
wwadman: Removed

src/power_grid_model_ds/_core/model/grids/_helpers.py
```diff
@@ -41,3 +58,65 @@ def create_empty_grid(grid_class: Type[G], graph_model: type[BaseGraphModel] = R
...
     empty_fields["graphs"] = GraphContainer.empty(graph_model=graph_model)
     return grid_class(**empty_fields)
+
+
+def merge_grids(grid: G, other_grid: G, mode: str) -> None:
+    """See Grid.merge()"""
+
+    match mode:
+        case "recalculate_ids":
+            other_grid = copy.deepcopy(other_grid)
```
jaapschoutenalliander: Afterthought about performance, this also copies graphs and other objects in the grid while these are not modified nor used (arrays are modified and appended only). So it would suffice to copy arrays only. Something like

```
other_arrays = copy.deepcopy(other_grid.all_arrays())
_increment_grid_ids_by_offset(other_arrays, offset)

for array in other_grid.all_arrays():
    grid.append(array, check_max_id=False)
```

src/power_grid_model_ds/_core/model/grids/base.py
```diff
@@ -379,6 +383,18 @@ def cache(self, cache_dir: Path, cache_name: str, compress: bool = True):
         )
         return save_grid_to_pickle(self, cache_dir=cache_dir, cache_name=cache_name, compress=compress)
 
+    def merge(self, other_grid: G, mode: str) -> None:
```
vincentkoppen: Lets make mode a `Literal`  at the minimal, to help with type hinting / typo errors. 

(ALso needs the literal import at the top)
```suggestion
    def merge(self, other_grid: G, mode: Literal["recalculate_ids", "keep_ids"]) -> None:
```

src/power_grid_model_ds/_core/model/grids/_helpers.py
```diff
@@ -41,3 +58,65 @@ def create_empty_grid(grid_class: Type[G], graph_model: type[BaseGraphModel] = R
...
+            case TransformerTapRegulatorArray():
+                columns = ["regulated_object"]
+            case BranchArray():
+                columns = ["from_node", "to_node", "feeder_node_id", "feeder_branch_id"]
+            case Branch3Array():
+                columns = ["node_1", "node_2", "node_3"]
+            case SymGenArray() | SymLoadArray() | SourceArray():
+                columns = ["node"]
+            case _:
+                raise NotImplementedError(f"The array of type {type(array)} is not implemented for appending")
```
vincentkoppen: We could maybe add an call for action? 
To signal that we are willing to think about a more general solution if this is needed by anyone? 

Something along the line of: 
```suggestion
                raise NotImplementedError(f"The array of type {type(array)} is not implemented for appending. Let us know if more general support is needed.")
```

tests/unit/model/grids/test_helpers.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_merge_two_grids(self):
+        grid1 = Grid.from_txt("S1 2", "S1 3 link", "3 4 transformer")
+        grid2 = Grid.from_txt("S11 12", "S11 13 link", "13 14 transformer")
+        grid1_size = grid1.node.size
+        grid2_size = grid2.node.size
+
+        grid1.merge(grid2, mode="keep_ids")
+        merged_grid_size = grid1.node.size
+
+        assert merged_grid_size == grid1_size + grid2_size, "Merged grid size should be the sum of both grids' sizes"
```
vincentkoppen: Also test that we didn't alter the ids. I'm also like more direct values if the example is small enough, so we could do something like below. 

So something like below? 
```suggestion
    def test_merge_two_grids(self):
        grid1 = Grid.from_txt("S1 2", "S1 3 link", "3 4 transformer")
        grid2 = Grid.from_txt("S11 12", "S11 13 link", "13 14 transformer")

        grid1.merge(grid2, mode="keep_ids")


        assert grid1.node.id.tolist() == [1,2,3,4,11,12,13], "Merged node ids should be equal to those of grid1 and grid2 combined"
```

tests/unit/model/grids/test_helpers.py
```diff
@@ -0,0 +1,53 @@
...
+        grid1.merge(grid2, mode="recalculate_ids")
+        assert grid1.check_ids() is None, "Asset ids are not unique after merging!"
+
+        # Check if from and to nodes are updated by checking that their values form the entire set of node ids:
+        assert set(grid1.branches.from_node).union(grid1.branches.to_node) == set(grid1.node.id), (
+            "All from and to nodes should form the entire set of node ids in the merged grid!"
+        )
+
+        # assert node in grid.source is updated by checking if the node column contains values that are all node ids:
+        assert set(grid1.source.node).issubset(grid1.node.id), "All source nodes should be valid node ids!"
```
vincentkoppen: nit: 
Same here, I would prefer hardcoded values since the example is small enough. This removes the complexity from the test. 

We now the grid_counter should be 501 (or maybe 502?)
So than it becomes like 
```suggestion
        grid1 = Grid.from_txt("S1 2", "S1 3 link", "3 4 transformer")
        grid2 = Grid.from_txt("S1 2", "S1 13 link", "13 14 transformer")
        source = SourceArray(id=[501], node=[1], status=[1], u_ref=[0.0])
        grid1.append(source)
        grid2.append(source)

        grid1.merge(grid2, mode="recalculate_ids")
        assert grid1.check_ids() is None, "Asset ids are not unique after merging!"

        expected_offset = 501
        assert grid1.node.id = [1, 2, 3, 4, expected_offset +1, expected_offset + 2, expected_offset + 13,expected_offset  + 14]
        assert grid1.link.from_node = [.....]
```

src/power_grid_model_ds/_core/model/grids/_helpers.py
```diff
@@ -41,3 +58,68 @@ def create_empty_grid(grid_class: Type[G], graph_model: type[BaseGraphModel] = R
...
+        case "keep_ids":
+            pass
+        case _:
+            raise NotImplementedError(f"Merge mode {mode} is not implemented")
+
+    # Append all arrays from the first grid to the second
+    for array in other_grid.all_arrays():
+        grid.append(array, check_max_id=False)
+    
+    if mode == "keep_ids":
```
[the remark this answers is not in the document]
wwadman: Thanks, good addition!

src/power_grid_model_ds/_core/model/grids/_helpers.py
jaapschoutenalliander: Nice, looks good to me

PR #138, opened 2026-01-14
draft: metadata support for columns

vincentkoppen:
For #136 we would like to have a programmable way to recognize id columns. 

I tried by setting the type, however I did not really get anywhere with this. Instead I went with the metadata route in this setup. 
It works, but not to sure whether I/we are comfortable with this implementation. Specifically since numpy has the following warning ;) 

https://numpy.org/doc/stable/reference/generated/numpy.dtype.metadata.html
 >Although used in certain projects, this feature was long undocumented and is not well supported. Some aspects of metadata propagation are expected to change in the future.

How it works: 
- We add support to annotate a NDArray within a dtype definition with a dict (or we can choose a different datastructure)
- We set the metadata provided in the dtype
- We can now access the metadata where we want. 

The `main.py` example code has the following output: 
```python
```
```
[('id', '<i4'), ('measured_object', '<i4'), ('u_sigma', '<f8', (3,)), ('u_measured', '<f8', (3,)), ('u_angle_measured', '<f8', (3,))] #the dtype
{'id': True} #metadata of "id" in dtype
```


vincentkoppen:
Picked up with a different solution for now

Issue #144, opened 2026-01-27
[BUG] Grid.__eq__ does not work for extended Grids


Thijss:
Apparently the `@dataclass` overwrites the __eq__ method on a class.
We hadn't foreseen that. Therefore, currently the __eq__ method is broken for  ExtendedGrids that inherit from the base Grid.

We're currently considering options to solve this issue

PR #148, opened 2026-02-06
fix: define __eq__ on all attributes of Grid and use default __eq__

vincentkoppen:
**Fixes issue**:  #144 

- Add `__eq__` on the BaseGraph (which gives the GraphContainer a `__eq__` since it is a dataclass)
- Fix the `__eq__` on fancyarray
- Remove the `__eq__` on Grid, since the default `dataclass`  method now just works.

### DIFF
--- a/src/power_grid_model_ds/_core/model/graphs/models/base.py
+++ b/src/power_grid_model_ds/_core/model/graphs/models/base.py
@@ -416,12 +416,13 @@ class BaseGraphModel(ABC):
     def __eq__(self, other: object) -> bool:
         """Check if two graph models are equal by comparing their branches and nodes."""
         if not isinstance(other, BaseGraphModel):
-            return NotImplemented
+            return False
 
-        my_branches = Counter((frozenset(branch) for branch in self.all_branches))
-        other_branches = Counter((frozenset(branch) for branch in other.all_branches))
         return (
-            my_branches == other_branches
-            and set(self.external_ids) == set(other.external_ids)
+            set(self.external_ids) == set(other.external_ids)
             and self.active_only == other.active_only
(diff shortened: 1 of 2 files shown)


src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -412,3 +412,16 @@ def _find_fundamental_cycles(self) -> list[list[int]]: ...
 
     @abstractmethod
     def _all_branches(self) -> Generator[tuple[int, int], None, None]: ...
+
+    def __eq__(self, other: object) -> bool:
+        """Check if two graph models are equal by comparing their branches and nodes."""
+        if not isinstance(other, BaseGraphModel):
+            return NotImplemented
```
jaapschoutenalliander: Could this not just be return False?

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -412,3 +412,16 @@ def _find_fundamental_cycles(self) -> list[list[int]]: ...
...
+
+        my_branches = Counter((frozenset(branch) for branch in self.all_branches))
+        other_branches = Counter((frozenset(branch) for branch in other.all_branches))
+        return (
+            my_branches == other_branches
```
jaapschoutenalliander: Consider putting the fastest check first for speed

jaapschoutenalliander (approved):

src/power_grid_model_ds/_core/model/graphs/models/base.py
vincentkoppen: done

PR #151, opened 2026-02-11
feat: delete three winding trafo on node deletion

Rubenscheffer:

- Deleting a node using the grid delete_node method will now also remove any connected three_winding_transformers in addition to the other branches.


Thijss (approved):
LGTM

PR #155, opened 2026-02-12
feat: iterate over shortest path

jaapschoutenalliander:
## Changes proposed in this PR include:
This pull request introduces a new feature for inspecting branch records along the shortest path enabling users to iterate through the branch objects that make up the shortest path between two nodes, rather than just the node sequence.

```python
    def iter_branches_in_shortest_path(
        self, from_node_id: int, to_node_id: int, typed: bool = False
    ) -> Iterator[BranchArray]:
        """Returns the ordered active branches that form the shortest path between two nodes.

        Args:
            from_node_id (int): External id of the path start node.
            to_node_id (int): External id of the path end node.
            typed (bool): If True, each yielded branch is converted to its typed array via
                ``get_typed_branches``.

        Yields:
            BranchArray: branch arrays for each active branch on the path.

        Raises:
            MissingBranchError: If the graph reports an edge on the shortest path but no active branch is found.
        """
```

## For discussion
- How to handle three-winding-transformers in the path

ToDo
- [ ] Add some tests (@vincentkoppen)


src/power_grid_model_ds/_core/model/grids/_search.py
```diff
@@ -71,3 +73,82 @@ def get_downstream_nodes(grid: "Grid", node_id: int, inclusive: bool = False):
...
+
+
+_THREE_WINDING_BRANCH_CONFIGS = (
+    ("node_1", "node_2", "status_1", "status_2"),
+    ("node_1", "node_3", "status_1", "status_3"),
+    ("node_2", "node_3", "status_2", "status_3"),
+)
+
+
+def _lookup_three_winding_branch(grid: "Grid", node_a: int, node_b: int) -> ThreeWindingTransformerArray:
```
Thijss: I'd prefer having both `_private` function below the public one.
(This is also recommended in the book Clean Clode)

Thijss (dismissed):
1 comment, but LGTM

PR #157, opened 2026-02-15
docs: setup feeder ids notebook

jaapschoutenalliander:
Add documentation on the feeder ids functionality

### DIFF
--- a/docs/examples/model/feeder_ids_examples.ipynb
+++ b/docs/examples/model/feeder_ids_examples.ipynb
@@ -25,7 +25,7 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


docs/examples/model/feeder_ids_examples.ipynb
vincentkoppen: Inspecting results section
Also note that all substation nodes have an empty feeder_node_id/feeder_branch_id value?

vincentkoppen: Mention that if you have a route with two substations, then any of the the feeder branches is chosen for the whole route?

PR #176, opened 2026-03-05
Add grid.diff() method and diff utils

Thijss:
Adds a Grid.diff() debug method that prints all the differences between two grids.

Also exposes the underlying `power_grid_model_ds.utils.find_differences_between_grids` and `power_grid_model_ds.utils.find_diff_masks_with_equal_nan` for custom analyses

Example usage:
```
>>> grid1 = Grid.from_txt("1 2 10", "2 3 11")
>>> grid2 = Grid.from_txt("1 2 10", "2 4 11")
>>> grid1.diff(grid2)

Difference in 'grid.graphs':
----------------------------
GraphContainer(active_graph=RustworkxGraphModel(nodes=3, branches=2, active_only=True), complete_graph=RustworkxGraphModel(nodes=3, branches=2, active_only=False))

vs.

GraphContainer(active_graph=RustworkxGraphModel(nodes=3, branches=2, active_only=True), complete_graph=RustworkxGraphModel(nodes=3, branches=2, active_only=False))

*** Note: Differences in graphs may not be visible in the representations above. Check the branch arrays below for actual differences. ***

Difference in 'grid.node':
--------------------------
 id | u_rated | node_type | feeder_branch_id | feeder_node_id 
 3  |   nan   |     0     |   -2147483648    |  -2147483648   

vs.

 id | u_rated | node_type | feeder_branch_id | feeder_node_id 
 4  |   nan   |     0     |   -2147483648    |  -2147483648   

Difference in 'grid.line':
--------------------------
 id | from_node | to_node | from_status | to_status | feeder_branch_id | feeder_node_id | is_feeder |  r1 |  x1 |  c1 | tan1 | i_n 
 11 |     2     |    3    |      1      |     1     |   -2147483648    |  -2147483648   |   False   | nan | nan | nan | nan  | nan 

vs.

 id | from_node | to_node | from_status | to_status | feeder_branch_id | feeder_node_id | is_feeder |  r1 |  x1 |  c1 | tan1 | i_n 
 11 |     2     |    4    |      1      |     1     |   -2147483648    |  -2147483648   |   False   | nan | nan | nan | nan  | nan 

```

### DIFF
--- a/src/power_grid_model_ds/experimental.py
+++ b/src/power_grid_model_ds/experimental.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: MPL-2.0
 """This module exposes functions/classes that are still under development and potentially subject to change."""
+
 from power_grid_model_ds._core.model.grids._search import find_differences_between_grids
 from power_grid_model_ds._core.utils.misc import find_diff_masks_with_equal_nan
 


src/power_grid_model_ds/_core/model/grids/_search.py
```diff
@@ -71,3 +74,48 @@ def get_downstream_nodes(grid: "Grid", node_id: int, inclusive: bool = False):
     return grid.graphs.active_graph.get_downstream_nodes(
         node_id=node_id, start_node_ids=list(substation_nodes.id), inclusive=inclusive
     )
+
+
+def find_differences_between_grids(
+    grid1: "Grid", grid2: "Grid", print_diff: bool = False
```
mgovers: would it be possible to also support plain PGM grids (with raw numpy arrays)?

mgovers: indeed. plain PGM grids are of the form `dict[str, np.ndarray] | dict[str, dict[str, np.ndarray]]` (the latter is for columnar data), but `FancyArray` is pgm-ds specific.

Thijss: I'm not sure pgm-ds is the right place for that.

Within pgm-ds is it assumed you use a pgm-ds grid.

Seems to me that a pgm dict comparison tool should live within PGM itself.

We could opt to move the  find_diff_masks_with_equal_nan function to PGM, so we can add the tool there.

Alternatively, it only takes one line of code to convert a pgm dict to a pgm-ds grid. So if you want to use extra functionalities like this, why not just convert to a pgm-ds grid?

src/power_grid_model_ds/_core/model/grids/_search.py
```diff
@@ -71,3 +74,51 @@ def get_downstream_nodes(grid: "Grid", node_id: int, inclusive: bool = False):
...
+
+def _compare_attr(attr1: object, attr2: object) -> dict[str, object]:
+    if isinstance(attr1, FancyArray) and isinstance(attr2, FancyArray):
+        if array_equal(attr1, attr2):
+            return {}
+        mask1, mask2 = find_diff_masks_with_equal_nan(attr1.data, attr2.data)
+        if mask1.any() or mask2.any():
+            return {"grid1": attr1[mask1], "grid2": attr2[mask2]}
+    if attr1 != attr2:
+        return {"grid1": attr1, "grid2": attr2}
```
vincentkoppen: There is a possible path of checking an array twice.
It should occur, but might be nice to circumvent? 
- `array_equal` returns `False`, but `find_diff_masks_with_equal_nan` doesn't return any mask. Then we do a seperate `attr1 != attr2` on the arrays again.

src/power_grid_model_ds/_core/model/grids/_search.py
```diff
@@ -71,3 +74,51 @@ def get_downstream_nodes(grid: "Grid", node_id: int, inclusive: bool = False):
...
+    return differences
+
+
+def _compare_attr(attr1: object, attr2: object) -> dict[str, object]:
+    if isinstance(attr1, FancyArray) and isinstance(attr2, FancyArray):
+        if array_equal(attr1, attr2):
+            return {}
+        mask1, mask2 = find_diff_masks_with_equal_nan(attr1.data, attr2.data)
+        if mask1.any() or mask2.any():
+            return {"grid1": attr1[mask1], "grid2": attr2[mask2]}
```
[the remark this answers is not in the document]
Thijss: https://github.com/PowerGridModel/power-grid-model-ds/issues/178

src/power_grid_model_ds/experimental.py
```diff
@@ -0,0 +1,8 @@
+# SPDX-FileCopyrightText: Contributors to the Power Grid Model project <powergridmodel@lfenergy.org>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+from power_grid_model_ds._core.model.grids._search import find_differences_between_grids
+from power_grid_model_ds._core.utils.misc import find_diff_masks_with_equal_nan
+
+__all__ = ["find_differences_between_grids", "find_diff_masks_with_equal_nan"]
```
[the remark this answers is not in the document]
Thijss: Additionally, these two functions do give return values, so they could be used in custom analysis scripts

PR #177, opened 2026-03-06
Add grid.rebuild_graphs()

Thijss:
The `.from_arrays` method on the `GraphContainer` has a bit of a confusing name.
Also, it is probably easier for end users to use a function directly on the Grid.

This PR deprecates the old methods and introduces the new `Grid.rebuild_graphs()` method

### DIFF
--- a/src/power_grid_model_ds/_core/model/grids/base.py
+++ b/src/power_grid_model_ds/_core/model/grids/base.py
@@ -464,7 +464,7 @@ class Grid(FancyArrayContainer):
     def rebuild_graphs(self) -> None:
         """(Re)build the graphs in the grid."""
         self.graphs = GraphContainer.from_grid(self)
-    
+
     def diff(self, other_grid: Self) -> None:
         """Print the differences between two grids
 


src/power_grid_model_ds/_core/model/graphs/container.py
```diff
@@ -120,12 +121,23 @@ def make_inactive(self, branch: BranchArray) -> None:
...
+            "Use Grid.rebuild_graphs() instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return cls._from_grid(arrays)
+
+    @classmethod
+    def _from_grid(cls, grid: "Grid") -> "GraphContainer":
+        """Build from grid"""
+        cls._validate_branches(arrays=grid)
```
vincentkoppen: Same preference here

vincentkoppen (commented):
Agree that `rebuild_graph`  is a nice addition, see the remarks regarding the deprecation.

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -349,12 +349,21 @@ def find_fundamental_cycles(self) -> list[list[int]]:
...
-
-        new_graph.add_node_array(node_array=arrays.node, raise_on_fail=False)
-        new_graph.add_branch_array(arrays.branches)
-        new_graph.add_branch3_array(arrays.three_winding_transformer)
+        warnings.warn(
+            f"{cls.__name__}.from_arrays is deprecated and will be removed in a future release. "
+            f"Use {cls.__name__}.from_grid instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
vincentkoppen: nit: if we deprecate, let also mention it in the docstrings (this is what you see on a hover in an editor)
As said before: I'm fine with not deprecating it yet and just keeping it as the symlink it currently is.
But I don't assume many users to use the function directly in a lot of places, so updating should be to much of an hassle.

vincentkoppen (approved):
See comments, but I'm with merging either way.

PR #197, opened 2026-03-20
feat(visualizer): Attached components visualization

nitbharambe:
Fixes #issue_number

- Reroute element selection via associated_ids
- Parse remaining components

Piggybacks:

- Element selection is now also possible for multi selection via ctrl + click and by a box selection.
- min max zoom restriction. This is something i faced as my mouse has a smooth scroll and the grid gets too zoomed out.

https://github.com/user-attachments/assets/1e2fa9af-4aac-4dab-af23-f3f9f18bfc90

## Follow ups

- Box selection and possibility of all element visualization is to be documented in #257.
- Another follow up is #259. We should also provide users a way to disable certain options.


src/power_grid_model_ds/_core/visualizer/callbacks/element_selection.py
```diff
@@ -24,21 +23,32 @@
 )
 def display_selected_element(node_data: list[dict[str, Any]], edge_data: list[dict[str, Any]]):
     """Display the tapped edge data."""
-    # 0th element means data for only a single selection is shown
-    if node_data:
-        selected_data = node_data.pop()
-    elif edge_data:
-        selected_data = edge_data.pop()
-    else:
+    if not node_data and not edge_data:
```
nitbharambe, line 26: A piggyback of multi-selection. Instead of one element, make table for all selected elements

Thijss: While I like the addition 👍 , next time I think this would deserve a separate PR.

tests/unit/visualizer/test_parsers.py
```diff
@@ -399,3 +407,113 @@ def test_parsing_with_starting_elements(func, request, test_data, kwargs):
 
     assert len(elements) == len(expected_elements)
     assert elements == expected_elements
+
+
+def test_parse_flow_sensors() -> None:
```
Thijss, line 412: Personally I prefer only testing public functions where possible. In this case I think we could also create a simple Grid with a two nodes and a branch + a sensor. Then we can just test `parse_element_data()` instead.

Thijss (commented):
Nice addition! I have some minor remarks, but overall looking good.

tests/unit/visualizer/test_parsers.py
```diff
@@ -399,3 +407,113 @@ def test_parsing_with_starting_elements(func, request, test_data, kwargs):
...
+    expected_elements["1"]["data"]["associated_ids"].update({ComponentType.sym_power_sensor.value: [1001]})
+    expected_elements["102_0"]["data"]["associated_ids"].update({ComponentType.sym_power_sensor.value: [1002]})
+    expected_elements["102_1"]["data"]["associated_ids"].update({ComponentType.sym_power_sensor.value: [1002]})
+    expected_elements["102_2"]["data"]["associated_ids"].update({ComponentType.sym_power_sensor.value: [1002]})
+
+    assert len(elements) == len(expected_elements)
+    assert elements == expected_elements
+
+
+def test_parse_flow_sensors_invalid_measured_terminal_type() -> None:
```
Thijss, line 510: same as https://github.com/PowerGridModel/power-grid-model-ds/pull/197/changes#r3235176778

src/power_grid_model_ds/_core/visualizer/callbacks/element_selection.py
jaapschoutenalliander: Maybe also good to document this feature. We can add a comment in the docs and maybe in https://github.com/PowerGridModel/power-grid-model-ds/blob/c3afcdb6342f75802a3392709576609c069abd26/src/power_grid_model_ds/_core/visualizer/layout/selection_output.py#L10

src/power_grid_model_ds/_core/visualizer/callbacks/element_selection.py
```diff
@@ -24,21 +23,32 @@
...
+    tables: list[html.H5 | dag.AgGrid] = []
+    for comp_type, ids in items_to_visualize.items():
+        # Select unique as multiple selected elements can be connected to the same component
+        # (e.g. appliance and node selected together, branch3 edges selected together)
+        unique_ids = list(set(ids))
+        data = getattr(grid, comp_type).filter(id=unique_ids)
+        tables.append(html.H5(comp_type, style={"marginTop": "15px", "textAlign": "left"}))
+        tables.append(_to_data_table(data))
+
+    return html.Div(children=tables, style={"overflowX": "scroll", "margin": "10px"}).children
```
jaapschoutenalliander: Could this simplify to `return tables`? What does the html.Div add if we then take out the tables only

src/power_grid_model_ds/_core/visualizer/callbacks/element_selection.py
jaapschoutenalliander: Also, maybe better to rename to `display_selected_elements` in plural

jaapschoutenalliander: What happens when you select a 1000 nodes in a large graph?

jaapschoutenalliander (commented):
Very cool features

src/power_grid_model_ds/_core/visualizer/callbacks/element_selection.py
nitbharambe: Indeed. Changed.

tests/unit/visualizer/test_parsers.py
nitbharambe: That is indeed possible. 
It was earlier unitized and I tired to keep it as such. But it was only 3 functions before!

2 nodes a branch and sensor would not cover all elements right?
I think the fixture `topologically_full_grid` can fully cover all code of parsers. Do we switch to that instead?

src/power_grid_model_ds/_core/visualizer/callbacks/element_selection.py
nitbharambe: >  I think this would deserve a separate PR.

Definitely. I think I added it because it was avoiding some roundabout changes if made into separate PR. 

> Maybe also good to document this feature. 

Renamed and documented `Click on a **node** or **edge** to display all its associated components and their attributes. You can also use Ctrl+Click (or Cmd+Click on Mac) to select multiple nodes or edges.`

> What happens when you select a 1000 nodes in a large graph?

For large selection of nodes:
Searched the internet and i think maybe the selection process happens this way:
- We first convert to the grid arrays to python form with `_to_data_table`. 
- Then dash serializes this to JSON. 
- Then the data is deserialized and displayed on demand (At the moment when we reach bottom while scrolling and it displays new data.) . 
 
I guess the first step would be expensive and storage of the serialized data in browser would require quite a bit of memory. All in all this seems to be maybe <10x computation and memory as parsing of elements. 

TLDR: Same level of consequence as visualizing large grid.

tests/unit/visualizer/test_parsers.py
nitbharambe: To be handled in #258

tests/unit/visualizer/test_parsers.py
nitbharambe: To be handled in #258

Thijss (approved):

PR #198, opened 2026-03-20
fix: bugfix three winding transformers in graphs

### LINKED ISSUE #167: [BUG] Three winding transformers are represented as cycles in the graph
**Describe the bug**
The support within the graphs for three windings arrays has some strange side effect currently for some of the graphs functions. 

Below shows the identified problems
```
   grid = Grid.empty()

    node_array = NodeArray.empty(3)
    node_array.id = [1, 2, 3]
    grid.append(node_array)

    three_winding_array = ThreeWindingTransformerArray.empty(1)
    three_winding_array.node_1 = 1
    three_winding_array.node_2 = 2
    three_winding_array.node_3 = 3
    three_winding_array.status_1 = 1
    three_winding_array.status_2 = 1
    three_winding_array.status_3 = 1
    grid.append(three_winding_array)

    cycles = grid.graphs.active_graph.find_fundamental_cycles()
    # cycles = [[2, 1, 3, 2]]
    paths = grid.graphs.active_graph.get_all_paths(1, 2)
    # paths = [[1,3,2], [1,2]]
```

vincentkoppen:
Depends on #267  to be merged first
Fixes #167 

There were multiple ways to solve this problem: 
- Keep 2 representations of the graph internally and use the "correct" one for each algorithm
Disadvantage on memory
- Represent a three-winding transformer as a star and adjust algorithms where necessary.
Disadvantage: a lot of current algorithms/queries become harder, since you always need to hop / remove the node.
- Keep representation as is (a cycle), but fix the algorithms that are hit by the problem (currently only 2 that share a lot of similarity)

Neither option is perfect, but for now keeping it mostly as is and fixing the two algorithms that were incorrect seems like the easiest/cleanest option.
For these algorithms we: 
- Remove one edge of the cycle (if all three active)
- Perform normal algorithm
- If any of the returned paths contains all three nodes of the three winding transformer after each other, we know that in reality we would have taken the removed edge instead of "walking through the three winding trafo". So we replace those
- Return to user
This way we do not return additional paths that are not there, nor do we see cycles that are not there (since we removed the cycles introduced by the three winding trafo's). 

Later if we add even more algorithm we can always revisit again en see whether one of the other option becomes a better solution.


src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -250,19 +278,68 @@ def get_shortest_path(self, ext_start_node_id: int, ext_end_node_id: int) -> tup
...
         Returns a list of paths, each path containing a list of external nodes.
         """
         if ext_start_node_id == ext_end_node_id:
             return []
 
-        internal_paths = self._get_all_paths(
-            source=self.external_to_internal(ext_start_node_id),
-            target=self.external_to_internal(ext_end_node_id),
-        )
+        branches_to_remove = self._branches_to_remove_from_three_winding_transformers()
```
jaapschoutenalliander: I suggest putting the boiler plate to remove the three winding transformers and re-add them in a re-usable wrapper. If there are more algorithms that need it they can easily do that:

something like
```python
    def _run_with_three_winding_correction(
        self,
        algorithm: Callable[[], list[list[int]]],
    ) -> list[list[int]]:
        branches_to_remove = self._branches_to_remove_from_three_winding_transformers()
        with self.tmp_remove_branches(branches_to_remove):
            result = algorithm()
        if branches_to_remove:
            return self._squash_paths_inside_three_winding_transformers(result)
        return result
```

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -223,6 +232,25 @@ def tmp_remove_nodes(self, nodes: list[int]) -> Generator:
...
+            >>> assert graph.has_branch(1, 2)
+            >>> assert graph.has_branch(2, 3)
+        """
+        for from_node, to_node in branches:
+            self.delete_branch(from_node, to_node)
+
+        yield
+
+        for from_node, to_node in branches:
+            self.add_branch(from_node, to_node)
```
jaapschoutenalliander: Also goes for the existing tmp_remove_nodes, if the algorithm inside the with block raises, the graph is left permanently mutated. Wrapping the yield in try/finally fixes it.

```python
try:
    yield
finally:
    for from_node, to_node in branches:
        self.add_branch(from_node, to_node)
```

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -200,6 +208,7 @@ def delete_branch3_array(self, branch3_array: Branch3Array, raise_on_fail: bool
         """Delete all branch3s in the branch3 array from the graph."""
         for branch3 in branch3_array:
             self.delete_branch_array(branch3.as_branches(), raise_on_fail=raise_on_fail)
+            self._three_winding_nodes.remove(self._get_branch3_nodes(branch3))
```
jaapschoutenalliander: When raise_on_fail = False this might be reached without there actually being an element in the set.
```suggestion
            self._three_winding_nodes.discard(self._get_branch3_nodes(branch3))
```

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -384,6 +468,12 @@ def _branch_is_relevant(self, branch: BranchArray) -> bool:
             return branch.is_active.item()
         return True
 
+    def _get_branch3_nodes(self, branch3_array: Branch3Array) -> tuple[int, int, int]:
+        """Get the nodes of the branch3 array as a set of tuples"""
```
jaapschoutenalliander:
```suggestion
        """Get the node ids of the branch3 array as a tuple"""
```

jaapschoutenalliander (commented):
Happy with the suggested solution for now. Some input on the implementation but mainly looks good

src/power_grid_model_ds/_core/model/graphs/models/base.py
Thijss: Seems like a good idea. I guess for readability we could even make a contextmanager.

```python
with self._three_winding_correction():
    ...
```

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -223,6 +232,25 @@ def tmp_remove_nodes(self, nodes: list[int]) -> Generator:
         for source, target in edge_list:
             self.add_branch(source, target)
 
+    @contextmanager
+    def tmp_remove_branches(self, branches: list[tuple[int, int]]) -> Generator:
```
Thijss, line 236: As discussed with @vincentkoppen , probably better to add this first in a separate PR

src/power_grid_model_ds/_core/model/graphs/models/base.py
vincentkoppen: Nice find.
I made a PR (#266 ) to fix it  for `tmp_remove_nodes` and made a seperate PR to introduce this `tmp_remove_branches` functionality (#267 ) with the fix included

src/power_grid_model_ds/_core/model/graphs/models/base.py
vincentkoppen: Done: #267

src/power_grid_model_ds/_core/model/graphs/models/base.py
vincentkoppen: Nice, added a test for this case as well.

src/power_grid_model_ds/_core/model/graphs/models/base.py
vincentkoppen: I will experiment with this, for the current functionality this works. But only since the two algorithms that we need to "fix" happen to have the same output (`list[list[int]]`). If we have a third algorithm later on that needs fixing with a different output type, we might have to revisit. 
--> Think I can also solve this by also applying a "post-processs" function that we can now set as default to the current one

vincentkoppen: I don't think a context manager can work, unless we do not include the `_squash_paths_inside_three_winding_transformers` part in it. 
We do not get access to the result of the function that is called within the context manager, which is needed for that function.

vincentkoppen: @Thijss  and @jaapschoutenalliander  I have simplified the setup I think, let me know what you think! :)

jaapschoutenalliander: Nice, happy with this approach

jaapschoutenalliander (approved):

PR #223, opened 2026-04-10
feat: serialize to JSON string instead of file

jaapschoutenalliander:
This pull request adds support for serializing and deserializing grid objects to and from JSON strings (in addition to files)

fixes: #210


jaapschoutenalliander:
@TomInfiniot, we have set up a PR which gives to option to serialize to json_string. Does this also suit your needs?

src/power_grid_model_ds/_core/model/grids/base.py
```diff
@@ -464,20 +469,55 @@ def merge(self, other_grid, mode: Literal["keep_ids", "recalculate_ids"]):
...
-    def serialize(self, path: Path, **kwargs) -> Path:
+    @overload
+    def serialize(self, path: Path, mode: Literal["json"] = "json", **kwargs) -> Path: ...
+
+    @overload
+    def serialize(self, path: None = None, *, mode: Literal["json_string"], **kwargs) -> str: ...
+
+    def serialize(
+        self, path: Path | None = None, mode: Literal["json", "json_string"] = "json", **kwargs
+    ) -> Path | str:
```
Thijss: Is it necessary to add the union output here? I'm not sure but might potentially confuse typecheckers? Could be mistaken, but I thought typing the overload functions is enough.

jaapschoutenalliander: I think it is not needed indeed, following https://peps.python.org/pep-0484/#function-method-overloading i removed the typing on the output here

Thijss (approved):
LGTM

vincentkoppen (approved):
Looks good!

PR #225, opened 2026-04-13
Add explicit size property to FancyArray

### LINKED ISSUE #221: [BUG] Pyright has trouble infering the type of an array's .size property
**Describe the bug**

When calling a FancyArray's .size property, Pyright cannot correctly infer the type of the property (which should be int).

**To Reproduce**

Minimal examply script:
```python
from power_grid_model_ds.arrays import NodeArray

arr = NodeArray.zeros(5)
arr2 = NodeArray.zeros(arr.size)
```

Results in the following Pyright error: 
```
/home/tom/github/inpai/meuk.py:4:24 - error: Argument of type "ndarray[_AnyShape, dtype[Unknown]] | Any" cannot be assigned to parameter "num" of type "int" in function "zeros"
    Type "ndarray[_AnyShape, dtype[Unknown]] | Any" is not assignable to type "int"
      "ndarray[_AnyShape, dtype[Unknown]]" is not assignable to "int" (reportArgumentType)
```

Funnily enough, using `len(arr)` instead of `arr.size` does not produce an error.

**Expected behavior**

This should not give an error, since an Array's .size attribute should be of type int.

TomvdPoll:
Fixes #221 

- Add the size property explicitly to the FancyArray class, so that static type checkers can infer the property type

- Not sure what the testing strategy is for the repo, but I added a simple test to ensure .size works

### DIFF
--- a/src/power_grid_model_ds/_core/model/arrays/base/array.py
+++ b/src/power_grid_model_ds/_core/model/arrays/base/array.py
@@ -273,6 +273,10 @@ class FancyArray(ABC):  # noqa: B024
             return tpl_cls(*self._data)
         return tpl_cls(*self._data[0])
 
+    @property
+    def size(self) -> int:
+        return self._data.size
+
     def filter(
         self: Self,
         *args: int | Iterable[int] | np.ndarray,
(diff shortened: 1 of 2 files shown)


jaapschoutenalliander (approved):
LGTM, @vincentkoppen ok with this as well?

vincentkoppen (approved):

PR #235, opened 2026-04-21
[IMPROVEMENT] Update ruff rule: "PT"

furqan463:
Fixes #188 

Added ruff rule: "PT"


tests/conftest.py
```diff
@@ -37,18 +37,18 @@ def get_installed_graph_models() -> dict[str, type[BaseGraphModel]]:
 @pytest.fixture
 def grid() -> Generator[Grid, Any, None]:
     """A grid fixture that will be parametrized"""
-    yield Grid.empty()
+    yield Grid.empty()  # noqa: PT022
```
furqan463: replacing yield with return results in mypy failure regarding return type. Not sure about it, so added exception here and below.

tests/unit/model/arrays/test_build.py
```diff
@@ -80,7 +80,7 @@ def test_build_from_kwargs():
 
 
 def test_build_from_kwargs_with_missing_input_fields():
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match=r"Missing required columns: {'test_bool', 'test_str'}"):
```
vincentkoppen: Sets are random in order, so this can make the tests fail randomly. 
For now I would by happy enough with just matching on: `Missing required column` 

If we do want to check the columns, we could do it like this I suppose
```suggestion
    with pytest.raises(ValueError, match=r"Missing required columns") as error:
       ...
    error_message = str(error.value)
    assert "test_bool" in error_message
    assert "test_str" in error_message
```

tests/unit/model/arrays/test_build.py
```diff
@@ -219,20 +219,20 @@ def test_from_structured_subarray_with_defaults(fancy_test_array: FancyTestArray
 
 
 def test_from_structured_subarray_no_defaults(fancy_test_array: FancyTestArray):
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Missing required columns: {'test_float2', 'test_float3'}"):
```
vincentkoppen: Same here, the order is not guaranteed. So the test can randomly fail. 
The same proposed solutions apply

tests/unit/model/arrays/test_build.py
```diff
@@ -219,20 +219,20 @@ def test_from_structured_subarray_with_defaults(fancy_test_array: FancyTestArray
...
         ExtendedFancyTestArrayNoDefaults(fancy_test_array.data)
 
 
 def test_from_sub_ndarray_with_defaults(fancy_test_array: FancyTestArray):
     # defaults are not supported when working with unstructured arrays
     sub_ndarray = np.array(fancy_test_array.tolist())
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match=r"Cannot convert array of shape \(3, 5\) into 7 columns."):
```
vincentkoppen: nitpick: I prefer `re.escape` instead of manually escaping. 

```suggestion
    with pytest.raises(ValueError, match=re.escape("Cannot convert array of shape (3, 5) into 7 columns.")):
```

tests/unit/model/arrays/test_build.py
```diff
@@ -219,20 +219,20 @@ def test_from_structured_subarray_with_defaults(fancy_test_array: FancyTestArray
...
         ExtendedFancyTestArray(sub_ndarray)
 
 
 def test_from_sub_ndarray_no_defaults(fancy_test_array: FancyTestArray):
     sub_ndarray = np.array(fancy_test_array.tolist())
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match=r"Cannot convert array of shape \(3, 5\) into 7 columns."):
```
vincentkoppen: Same nitpick.

tests/unit/model/containers/test_base.py
```diff
@@ -38,14 +38,14 @@ class _FourArraysContainer(_TwoArraysContainer):
 
 
 def test_id_counter_type(basic_grid: Grid):
-    assert isinstance(basic_grid.id_counter, int)
+    assert isinstance(basic_grid.max_id, int)
```
vincentkoppen: While id_counter is deprecated, as long as we have it in our code I still want to test it. 
What was the reason that you changed this? All ruff checks seem to pass still if I change it back.

tests/unit/model/containers/test_base.py
```diff
@@ -38,14 +38,14 @@ class _FourArraysContainer(_TwoArraysContainer):
...
 
 
 def test_id_counter():
     container = FancyArrayContainer.empty()
     # pylint: disable=protected-access
     container._ids = {42}
-    assert container.id_counter == 42
+    assert container.max_id == 42
```
vincentkoppen: Same question here.

tests/unit/model/containers/test_base.py
```diff
@@ -210,7 +210,7 @@ def test_rebuild_ids():
 def test_rebuild_ids_with_duplicates():
     grid = Grid.from_txt("1 2 12")
     grid.node.id = [1, 12]  # Duplicate IDs within different arrays same array
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match=r"Duplicate ids found between arrays \(LineArray\)"):
```
vincentkoppen: Same nitpick about using `re.escape` instead of manually escaping.

tests/unit/model/grids/test_base.py
```diff
@@ -47,7 +47,7 @@ def test_from_extended_grid():
     assert grid.graphs.active_graph.nr_nodes == len(grid.node)
     assert grid.graphs.complete_graph.nr_nodes == len(grid.branches)
 
-    assert extended_grid.id_counter == grid.id_counter
+    assert extended_grid.max_id == grid.max_id
```
vincentkoppen: Same question regarding the change from `id_counter` to `max_id`.

tests/conftest.py
vincentkoppen: It's fine to change the return type to just `Grid` here. 
So it becomes: `def grid() -> Grid:`

tests/conftest.py
```diff
@@ -37,18 +37,18 @@ def get_installed_graph_models() -> dict[str, type[BaseGraphModel]]:
...
 
 
 @pytest.fixture
 def graph() -> Generator[RustworkxGraphModel, Any, None]:
     """A graph fixture that will be parametrized"""
-    yield RustworkxGraphModel()
+    yield RustworkxGraphModel()  # noqa: PT022
```
vincentkoppen: Same here, this should work: 

```python
@pytest.fixture
def graph() -> RustworkxGraphModel:
    """A graph fixture that will be parametrized"""
    return RustworkxGraphModel()
```

pyproject.toml
```diff
@@ -99,7 +99,7 @@ max-parents = 15
 max-line-length = 120
 
 [tool.ruff]
-output-format = "github"
+#output-format = "github"
```
vincentkoppen: Why did you comment this out?

vincentkoppen (changes requested):
Thanks again for the nice PR! 
Some questions and suggestions on this one.

pyproject.toml
furqan463: I did it for my taste and intended to remove at the end but forgot. will restore it.

tests/unit/model/containers/test_base.py
furqan463: It was giving deprecated warning, will reverse it.

tests/unit/model/arrays/test_build.py
furqan463: > If we do want to check the columns, we could do it like this I suppose

No this raises `PT012`.

furqan463: Fixed.

tests/unit/model/arrays/test_build.py
furqan463: Fixed.

tests/unit/model/arrays/test_build.py
furqan463: Fixed.

tests/unit/model/arrays/test_build.py
furqan463: Fixed.

tests/conftest.py
furqan463: Done.

tests/unit/model/containers/test_base.py
furqan463: Fixed.

tests/conftest.py
furqan463: Fixed.

tests/unit/model/containers/test_base.py
furqan463: Fixed.

tests/unit/model/containers/test_base.py
furqan463: fixed.

tests/unit/model/grids/test_base.py
furqan463: Fixed.

pyproject.toml
furqan463: fixed.

vincentkoppen (approved):
Looks good! 
One small merge conflict, but good to go apart from that.

Thijss (approved):

PR #236, opened 2026-04-24
feat: save visualizer html

jaapschoutenalliander:
- Added `save_html(grid, path)` to `visualizer.py` — exports the grid as a standalone HTML file without starting the Dash server
- New `html_export.py` module handles HTML generation: embeds graph elements, stylesheet, and layout as JSON with Cytoscape.js 
- Docs updated with usage example and parameter table; two unit tests cover default and non-default parameter paths


vincentkoppen (commented):
Looks like a nice addition. Haven't tested/viewed the HTML, but the code to maintain this doesn't seem to bad. 

Maybe some of the setup surround the layout/stylesheet can be shared even more with the actual app. But if not, this seems fine to me.

Thijss:
Looks promising! This has been requested for quite a while I believe 👍 . I'll review later this week

Thijss (approved):
I tested it locally. 

The HTML only contains the graph. The top and bottom sections are not included. In the future we could look whether that is feasible/desirable too (but we'd have to move logic to the JS side instead of Python).

For now, nice addition!

PR #241, opened 2026-05-04
feat: rewrite get_inherited_attrs (id-columns part 1/2)

Thijss:
PR in preparation of `_id_columns` PR (#243).

Splitted the logic around inherited attributes into two methods.

`build_mro_attribute` supports two types: `dict` and `set`

ToDo:
- [x] fix type issue

### DIFF
--- a/src/power_grid_model_ds/_core/utils/misc.py
+++ b/src/power_grid_model_ds/_core/utils/misc.py
@@ -5,7 +5,7 @@
 """Misc utils"""
 
 from collections.abc import Sequence
-from typing import get_type_hints, overload
+from typing import get_type_hints
 
 import numpy as np
 
@@ -30,15 +30,7 @@ def get_public_annotations(cls: type):
     return {attr: type_ for attr, type_ in class_attributes.items() if not attr.startswith("_")}
 
 
-@overload
-def build_mro_attribute(cls: type, attribute_name: str, attribute_type: type[dict]) -> dict: ...
-
-
-@overload
-def build_mro_attribute(cls: type, attribute_name: str, attribute_type: type[set]) -> set: ...
-
-
-def build_mro_attribute(cls: type, attribute_name: str, attribute_type) -> dict | set:
+def combine_attribute_from_parent_classes[T: (dict, set)](cls: type, attribute_name: str, attribute_type: type[T]) -> T:
     """Combine all versions of an attribute in the Method Resolution Order (mro) of a class into a single attribute
 
     For dicts this means the dict is updated so that child classes override parent classes.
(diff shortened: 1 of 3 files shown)


src/power_grid_model_ds/_core/utils/misc.py
```diff
@@ -23,22 +23,41 @@ def is_sequence(seq):
...
-        for parent in reversed(list(cls.__mro__)):
-            attr_dict = retrieved_attributes.get(private_attr, {})
-            attr_dict.update(getattr(parent, private_attr, {}))
-            retrieved_attributes[private_attr] = attr_dict
+@overload
+def build_mro_attribute(cls: type, attribute_name: str, attribute_type: type[set]) -> set: ...
 
-    return retrieved_attributes
+
+def build_mro_attribute(cls: type, attribute_name: str, attribute_type) -> dict | set:
```
vincentkoppen: We don't need overloading here. This is precisely what the 3.12 syntax made easy. 
Below should work/do the same. 

```
def build_mro_attribute[T: (dict, set)](cls: type, attribute_name: str, attribute_type: type[T]) -> T:
```

src/power_grid_model_ds/_core/utils/misc.py
```diff
@@ -23,22 +23,41 @@ def is_sequence(seq):
...
+
+    Types other than dict and set are not supported
+    """
+    attr_value = attribute_type()
+    for parent in reversed(list(cls.__mro__)):
+        if attribute_type is dict:
+            attr_value.update(getattr(parent, attribute_name, {}))
+        elif attribute_type is set:
+            attr_value |= getattr(parent, attribute_name, set())
+        else:
```
vincentkoppen, line 48: nit: 
you can do something like below to not repeat:
```python 
parent_attr = getattr(parent, attribute_name, attribute_type())
if attribute_type is dict: 
....
```

src/power_grid_model_ds/_core/utils/misc.py
```diff
@@ -23,22 +23,41 @@ def is_sequence(seq):
...
+
+def build_mro_attribute(cls: type, attribute_name: str, attribute_type) -> dict | set:
+    """Combine all versions of an attribute in the Method Resolution Order (mro) of a class into a single attribute
+
+    For dicts this means the dict is updated so that child classes override parent classes.
+    For sets this means the sets are unioned together.
+
+    Types other than dict and set are not supported
+    """
+    attr_value = attribute_type()
```
vincentkoppen: nit: `attr_value` -> `combined_attr`? 
The intent of this variable wasn't really clear to me straight away. This this might help.

src/power_grid_model_ds/_core/utils/misc.py
vincentkoppen: nit: naming: `combined_attribute_from_parent_classes`?
MRO is an order, so not clear what mro_attribute is to me. But maybe this is default naming? The proposed would be clearer to me atleast.

vincentkoppen (commented):
Looks like a good refactor + new functionality to me. 
Some nitpicks, but nothing major!

vincentkoppen (approved):
Looks good!

PR #243, opened 2026-05-05
feat: merge ids using get_id_columns (id-columns part 2/2)

Thijss:
Add _id_columns attribute to FancyArray, that is used during grid.merge to apply the offset to relevant columns in each array.

This allows grid.merge to also work with extended grids that contain arrays with id_columns not defined in PGM

### DIFF
--- a/src/power_grid_model_ds/_core/model/arrays/base/array.py
+++ b/src/power_grid_model_ds/_core/model/arrays/base/array.py
@@ -19,7 +19,11 @@ from power_grid_model_ds._core.model.arrays.base._optional import pd
 from power_grid_model_ds._core.model.arrays.base._string import convert_array_to_string
 from power_grid_model_ds._core.model.arrays.base.errors import ArrayDefinitionError
 from power_grid_model_ds._core.model.constants import EMPTY_ID, empty
-from power_grid_model_ds._core.utils.misc import array_equal_with_nan, build_mro_attribute, get_public_annotations
+from power_grid_model_ds._core.utils.misc import (
+    array_equal_with_nan,
+    combine_attribute_from_parent_classes,
+    get_public_annotations,
+)
 
 # pylint: disable=missing-function-docstring, too-many-public-methods
 
@@ -86,7 +90,7 @@ class FancyArray(ABC):  # noqa: B024
     @classmethod
     @lru_cache
     def get_defaults(cls) -> dict[str, Any]:
-        return build_mro_attribute(cls, "_defaults", attribute_type=dict)
+        return combine_attribute_from_parent_classes(cls, "_defaults", attribute_type=dict)
 
     @classmethod
     @lru_cache
@@ -97,7 +101,7 @@ class FancyArray(ABC):  # noqa: B024
     @lru_cache
     def get_dtype(cls):
         annotations = get_public_annotations(cls)
-        str_lengths = build_mro_attribute(cls, "_str_lengths", dict)
+        str_lengths = combine_attribute_from_parent_classes(cls, "_str_lengths", dict)
         dtypes = {}
         for name, dtype in annotations.items():
             if len(dtype.__args__) > 1:
--- a/src/power_grid_model_ds/_core/model/arrays/base/array.py
+++ b/src/power_grid_model_ds/_core/model/arrays/base/array.py
@@ -53,6 +53,10 @@ class FancyArray(ABC):  # noqa: B024
         The default length for string columns is stored in _DEFAULT_STR_LENGTH.
         To change this, you can set the _str_lengths class attribute.
 
+    Note on id-columns:
+        Columns that contain ids or reference ids (e.g., 'id', 'from_node', etc.) should be defined in the _id_columns
+        class attribute to ensure correct functionality of grid.merge
+
     Example:
         >>> class MyArray(FancyArray):
         >>>     name: NDArray[np.str_]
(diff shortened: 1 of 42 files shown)


src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -66,6 +71,7 @@ class FancyArray(ABC):  # noqa: B024
     _data: NDArray = np.ndarray([])
     _defaults: ClassVar[dict[str, Any]] = {}
     _str_lengths: ClassVar[dict[str, int]] = {}
+    _id_columns: ClassVar[set[str]] = set()
```
vincentkoppen, line 78: Do we introduce this in the FancyArray (here) or in `IdArray`? 

Against keeping it here FancyArray: 
A `id` is nothing in the context of an FancyArray, it only has meaning within the grid context. So this makes the FancyArray more thight to our grid. 

Against moving this logic to IdArray:
There might be use cases for users where they have an additional array in the grid that does not have an `id` column, but does reference id's in other columns. These would then be missed by our logic.

I think that for the current implementation it is fine to add this to `FancyArray`, since we do not advertise the `FancyArray` on it's own. But curious what other think

vincentkoppen: An alternative solution that might be the cleanest (but also something new for our package) is to introduce a mixin.

```python 
class IdColumnsMixin: 
   _id_columns: ClassVar[set[str]]

   @classmethod
   @lru_cache
   def get_id_columns(cls) -> set[str]:
        return build_mro_attribute(cls, "_id_columns", attribute_type=set)
```

We could then have `class IdArray(FancyArray, Id)` -> `class IdArray(FancyArray, Id, IdColumnsMixin)` and the user could also use the mixin for other non-IdArrays if needed.

src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -49,11 +49,16 @@ class FancyArray(ABC):  # noqa: B024
...
-    Note on string-columns:
+    Note on string columns:
         The default length for string columns is stored in _DEFAULT_STR_LENGTH.
         To change this, you can set the _str_lengths class attribute.
         The _str_lengths attribute is inherited by child classes.
 
+    Note on id columns:
+        Columns that contain ids or reference ids (e.g., 'id', 'from_node', etc.) should be defined in the _id_columns
+        class attribute to ensure correct functionality of grid.merge()
+        The _id_columns attribute is inherited by child classes.
```
vincentkoppen:
```suggestion
        Columns that contain ids or reference ids (e.g., 'id', 'from_node', etc.) should be defined in the _id_columns
        class attribute to ensure correct functionality of certain grid helper methods (e.g. grid.merge).
        The _id_columns attribute is inherited and combined with those of the parent classes.
```

src/power_grid_model_ds/_core/model/grids/_helpers.py
```diff
@@ -104,42 +83,10 @@ def merge_grids(grid, other_grid, mode: Literal["keep_ids", "recalculate_ids"]):
...
-            case FaultArray():
-                columns = ["fault_object"]
-            case _:
-                raise NotImplementedError(
-                    f"The array of type {type(array)} is not implemented for appending. "
-                    f"Let us know if more general support is needed."
-                )
-
-        for column in columns:
+        for column in array.get_id_columns():
```
vincentkoppen, line 86: 😍

tests/fixtures/grid_classes.py
```diff
@@ -3,17 +3,33 @@
...
+import numpy as np
+from numpy._typing import NDArray
 
 from power_grid_model_ds._core.model.grids.base import Grid
+from power_grid_model_ds.arrays import IdArray
 from tests.fixtures.arrays import CustomLineArray, CustomNodeArray, DefaultedCustomLineArray, DefaultedCustomNodeArray
 
 
+class ExtraArray(IdArray):
+    """An extra array with referencing id columns."""
```
vincentkoppen: nitpick: To make it a bit clearer to me. (No need for `id` column even, to make it as seperate as possible)

```suggestion
class UserDefinedArray(FancyArray):
    """An user defined array that does not inherit from any existing grid array with id references."""
```

src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -82,6 +88,11 @@ def data(self) -> NDArray:
     def get_defaults(cls) -> dict[str, Any]:
         return build_mro_attribute(cls, "_defaults", attribute_type=dict)
 
+    @classmethod
+    @lru_cache
+    def get_id_columns(cls) -> set[str]:
```
vincentkoppen, line 97: I don't see any test on this method.
Would be nice to just have a few even though it is very similar to `get_defaults`.

vincentkoppen (commented):
Looks good to me, some questions but in general no big remarks! 
This PR now changes our `merge`, but this could also greatly simplify our `delete_...` methods. But I would propose to change these in a seperate PR after this one.

src/power_grid_model_ds/_core/model/arrays/base/array.py
vincentkoppen: We have discussed it and for now we will keep it on the FancyArray. Mixin adds another inheritance layer with lirttle gains at this point. Also FancyArray is closely linked to the `Grid` anyways now.

Thijss: As discussed with @vincentkoppen, we'll stick with current implementation for now

src/power_grid_model_ds/_core/model/arrays/base/array.py
Thijss: will add

vincentkoppen (approved):

PR #244, opened 2026-05-05
chore: fix deprecation warnings from .id_counter usage in tests

Thijss:


vincentkoppen (commented):
I would say that we need to remove the `id_counter` as well if we remove the tests. 
Which I'm not opposed to.

Thijss:
Closed in favor of #254

PR #245, opened 2026-05-07
fix: Improve logic around grid.get_branches() for better performance

### LINKED ISSUE #238: [IMPROVEMENT] concatenate once for better performance
**Describe the problem**

get_branches has a pattern where it goes

array = empty array
for additional_array in iterator:
    array = concat(array, additional_array)

this causes a new array with new memory to be made a number of times (inefficient)

**Improvement proposal**
Prepping the arrays so we can do 1 concatenate will result in a massive speedup.

Proposed code
```

def get_branches(grid: "Grid") -> BranchArray:
    """see Grid.get_branches()"""
    branch_dtype = BranchArray.get_dtype()
    branch_columns = list(branch_dtype.names)
    branch_arrays = get_branch_arrays(grid)
    consistent_branch_arrays = [array[branch_columns] for array in branch_arrays if array.size]
    if len(consistent_branch_arrays) == 0:
        return BranchArray()
    return fp.concatenate(BranchArray(), *consistent_branch_arrays)
```

**Experimental results below**
With 10 'branches' the difference in performance is about 7-8x.
```
import numpy as np
import time

a = [np.zeros(1000) for _ in range(10)]
t_start = time.perf_counter()
for _ in range(100):
    new = a[0]
    for array in a:
        new = np.concatenate([new, array])
t_end = time.perf_counter()
print(f"concatenate: {t_end - t_start:0.8f} <- each time")


t_start = time.perf_counter()
for _ in range(100):
    new = np.concatenate(a)
t_end = time.perf_counter()
print(f"concatenate: {t_end - t_start:0.8f} <- once")
```

JohanMeppelinkDRA:
Fixes #238

Concatenate once is faster. So we now prepare the arrays in the list and concatenate once instead of concatenate each individually. This will help speed-up the get_branches function.

### DIFF
--- a/src/power_grid_model_ds/_core/model/grids/_search.py
+++ b/src/power_grid_model_ds/_core/model/grids/_search.py
@@ -22,11 +22,12 @@ if TYPE_CHECKING:
 def get_branches(grid: "Grid") -> BranchArray:
     """see Grid.get_branches()"""
     branch_dtype = BranchArray.get_dtype()
-    branches = BranchArray()
-    for array in grid.branch_arrays:
-        new_branch = BranchArray(data=array.data[list(branch_dtype.names)])
-        branches = fp.concatenate(branches, new_branch)
-    return branches
+    branch_columns = list(branch_dtype.names)
+    branch_arrays = get_branch_arrays(grid)
+    consistent_branch_arrays = [array[branch_columns] for array in branch_arrays if array.size]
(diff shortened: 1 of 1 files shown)


Thijss (approved):
Looks good to me

vincentkoppen (approved):
Nice improvement, thanks! 
Looks good to me as well.

PR #262, opened 2026-05-26
Add poe command 'test_notebooks'

Thijss:
I was struggling to debug issues with failing readthedocs builds.

This PR intends to simplify the debugging in the future

Adds new command:
- `poe test_notebooks`

This will run all notebooks. On failure a traceback is provided, otherwise a success message is printed

### Piggyback
renamed `clean_notebook` to `clean_notebooks` for consistency

### Open question

For now, the command is also added to `poe all` 
Do we want to do this? It adds a bit of duplication to the github CI/CD (two workflows then test the documentation flow), but I think it might be convenient to just add it.


jaapschoutenalliander (approved):
Nice, makes it very easy. If we add it to the all command it will impact the running time, is that acceptable? Tested it and in my opinion the running time is fine

vincentkoppen (approved):
Looks good to me as well, also fine with the extra time for the `poe all` command

PR #269, opened 2026-06-01
Implement `AttributeType` for `power_grid_model` objects

furqan463:
Fixes #206 

User `AttributeType` enum when interacting directly with `power_grid_model` objects/ interface only.

I have gone through whole code-base, test, docs. This should close the issue.


tests/unit/visualizer/test_grid_utils.py
```diff
@@ -85,39 +85,39 @@ def test_dynamic_grid_obj_from_grid(dataset_type):
...
 def test_get_attr_data_from_dataset():
     loads = initialize_array(DatasetType.update, ComponentType.sym_load, (4, 3))
-    loads["p_specified"] = [[21, 22, 23], [31, 32, 33], [41, 42, 43], [51, 52, 53]]
+    loads[AttributeType.p_specified] = [[21, 22, 23], [31, 32, 33], [41, 42, 43], [51, 52, 53]]
 
     # Normal case with matching pgm_id=12
-    loads["id"] = [[11, 12, 13], [11, 12, 13], [11, 12, 13], [11, 12, 13]]
+    loads[AttributeType.id] = [[11, 12, 13], [11, 12, 13], [11, 12, 13], [11, 12, 13]]
     batch_dataset = {ComponentType.sym_load: loads}
     x_actual, y_actual = get_attr_data_from_dataset(batch_dataset, ComponentType.sym_load, "p_specified", pgm_id=12)
```
furqan463, line 93: `loads` is directly initialised to pgm array, so `AttributeType` is used to access fields, while `get_attr_data_from_dataset` is pgm-ds method, so enum is not applied here.

vincentkoppen: Since this function only deals with PGM related objects, I would be fine to use `AttributeType` here as well. 
We could even update the function it self to: 
- change `attr: str` -> `attr: str | AttributeType`
- use `AttributeType.id` instead of `"id"` in the function
However if we do this, we should also update our minimal dependency on `power-grid-model` to the first version that introduced `AttributeType`. @jaapschoutenalliander  what do you think? 

We could always merge this PR as is and keep this change for a later time.

vincentkoppen (commented):
Thanks for the PR, looks good to me! 
One remark I would like to discuss with the other maintainers, we will get back to you on that.

tests/unit/visualizer/test_grid_utils.py
furqan463: > We could always merge this PR as is and keep this change for a later time.

That's better.

vincentkoppen (approved):

Issue #161, opened 2026-02-20
[FEATURE] Provide more basic graph tools


mgovers:
**Describe the feature request**

Currently, the graph representation only contains some more global tools (e.g.: get all cycles). However, it would be nice if there would be some tool to step locally through a graph (e.g.: get adjacent nodes/edges). The functionality already exists in the graph implementation, but is not yet exposed over the API and requires one to rebuild the graph from scratch.

vincentkoppen:
Thanks for the examples: 
- `adjacent(node:int) -> list[int]` and `adjacent_edges` definitely agree and open to PR's or might implement this myself when I have time
  - On: `adjacent_edges` will return a `list[tuple[int,int]]` since we don't have branch id's on the graph 
  - On: `adjacent(node: int, excluding: list[int]) -> list[int]`: seems like a nice optional argument indeed. Nit: would change the excluding to either be `set[int]` (for performance) or `Containable[int]` to support any structure that support the `x in excluded_list` check. 
- Since we don't have branch ids on the graph, currently `other_node` doesn't seem useful. Since a branch is just a tuple of the two nodes.  

The above also shows that we might want to think about supported `edge_id` (from our branches/three winding transformers) on our graphs. 

For `bfs` / `dfs` I agree that these are useful, will try to find some time in the upcoming weeks to see if I can get a grasp of what a good interface could look like.

juliettegiraud-2004:
Hello, I definitely agree with that request (all the functions adjacent, adjacent_edges, other_nodes, bfs, dfs would also be really useful for me)!  I would also suggest a couple of other complementary features that would be really useful for me as well : 

## Context: fault isolation and service restoration
My main use case is distribution network **fault isolation and restoration**.

In this context, I need to work with the same network at two different levels:

- a **full graph** for electrical computations (load flow, constraint checks),
- a **simplified topological graph**, where:
  - nodes represent *pockets / zones*,
  - edges represent *switches (breakers)* that can be opened/closed.

Each pocket corresponds to a subgraph of the full network, and the restoration logic is driven by the switches between these pockets.

A typical workflow is:
1. **Fault isolation**
   - open switches surrounding the faulty pocket
2. **Topological decomposition**
   - compute connected components on the simplified graph
3. **Classification**
   - run load flow
   - classify components into:
     - faulted (ignored),
     - energized,
     - non-energized or violating constraints
4. **Restoration search**
   - identify candidate open switches that connect an energized component to a non-energized one
[...]

PR #273, opened 2026-06-11
feat(graph): add adjacent method

vincentkoppen:
As part of #161 this implements the `adjacent` method as discussed there. 

A graph now has `.adjacent(node_id, excluding)` as a new method

### DIFF
--- a/src/power_grid_model_ds/_core/model/graphs/models/base.py
+++ b/src/power_grid_model_ds/_core/model/graphs/models/base.py
@@ -4,7 +4,7 @@
 import warnings
 from abc import ABC, abstractmethod
 from collections import Counter
-from collections.abc import Generator
+from collections.abc import Container, Generator
 from contextlib import contextmanager
 from itertools import combinations
 from typing import TYPE_CHECKING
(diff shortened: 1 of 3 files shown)


src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -103,6 +103,16 @@ def in_branches(self, node_id: int) -> Generator[tuple[int, int], None, None]:
             (self.internal_to_external(source), self.internal_to_external(target)) for source, target in internal_edges
         )
 
+    def adjacent(self, node_id: int, excluding: Container[int] | None = None) -> list[int]:
+        """Return all nodes connected to the given node.
+
+        Args:
+            excluding(Container[int]|None): exclude certain node ids frm the output of this function. Defaults to None.
```
jaapschoutenalliander, line 110: What is the use case for excluding nodes here?

jaapschoutenalliander (approved):

src/power_grid_model_ds/_core/model/graphs/models/base.py
vincentkoppen: See https://github.com/PowerGridModel/power-grid-model-ds/issues/161#issuecomment-3984814574
But basically I can image that you might want to "walk" through the graph without going back for instance. Then de excluding is helpful. since you can just keep track of every node you already visited and pass it to not make any loops/backwards movements.

PR #274, opened 2026-06-12
feat(graph): add bfs

vincentkoppen:
Part of #161 

(First) implementation of bfs, please give feedback if anything is missing

------
Updated: now returns a list of (node, parent)


tests/unit/model/graphs/test_graph_model.py
```diff
@@ -476,3 +476,26 @@ def test_has_parallel_edges_reversed(self, graph: BaseGraphModel):
...
+
+
+class TestBfsSearch:
+    @pytest.mark.parametrize(
+        ("source", "expected"),
+        [
+            pytest.param(1, [(1, None), (5, 1), (2, 1), (4, 5), (3, 2)], id="source: 1"),
+            pytest.param({1}, [(1, None), (5, 1), (2, 1), (4, 5), (3, 2)], id="source {1}"),
+            pytest.param([1, 2], [(1, None), (5, 1), (2, 1), (4, 5), (3, 2)], id="source [1,2]"),
+            pytest.param([2, 1], [(2, None), (3, 2), (1, 2), (5, 1), (4, 5)], id="source [1,2]"),
```
vincentkoppen:
```suggestion
            pytest.param([2, 1], [(2, None), (3, 2), (1, 2), (5, 1), (4, 5)], id="source [2,1]"),
```

src/power_grid_model_ds/_core/model/graphs/models/base.py
```diff
@@ -382,6 +382,25 @@ def get_downstream_nodes(self, node_id: int, start_node_ids: list[int], inclusiv
...
         return self.get_connected(node_id, [upstream_node], inclusive)
 
+    def bfs(self, source: int | Sequence[int]) -> list[tuple[int, int | None]]:
+        """Breadth first search from the source(s).
+
+        Args:
+            source(int | Sequence[int]): the source(s) the start the breadth first search from.
+
+        Returns:
+            list[tuple[int, int | None]]: the (node, parent) by the order in which we found the nodes.
```
jaapschoutenalliander: For this and the dfs PR: is it most natural to return (node, parent) or (parent, node)? For me the latter feels more easily to follow as you travel from parent to child

vincentkoppen: For me `(node, parent)` feels more natural, since the `node` on it's own might often be enough information; depending on your use case. While I don't see anyone only using the parents. 

The parent is metadata and helpful if you need more details on the exact path through the graph. 
Also you can now easily create a dict from the result if needed/preferred: 
`parents = dict(graph.bfs(...))` will give a node -> parent dict

vincentkoppen: From what I see in other libraries, they typically have multiple functions for different type of results. So one that returns the edges, one that returns the graph containing only the edges in the search path, etc.  
But that feels like a bit of overkill for us currently. It is however custom that the function that returns the edges is in the `(parent, node)` order it seems (but that wouldn't include the root, that we do include in the return)

We could also go for a dict return `{node: parent}`. In earlier versions of python the order of a dict wasn't guaranteed, but now it is based on insertion order. So looping over that dict will return the nodes in the correct order of the search. We would then still include the roots in this dictionary I guess, so you have: `{root: None, node_1: root, ...}`.

jaapschoutenalliander: I think these are two good options. Which do you prefer?

vincentkoppen: Discussed that we will go for `dict` return for now and ask at the meetup whether or not this is sufficient for everyone.

jaapschoutenalliander (approved):

PR #283, opened 2026-06-26
fix: support numpy v2.5

### LINKED ISSUE #282: [BUG] numpy 2.5 not supported
### Describe the bug

After installing `numpy` 2.5.0 I see that all test fail because of typing issues. 

The typing within our FancyArray's is no longer picked up correctly. 
``` ValueError: dtype NDArray[numpy.int32] not understood or supported```

### Input Data Validity

- [ ] I checked the validity of input data using the helper function ([examples here](https://power-grid-model.readthedocs.io/en/stable/examples/Validation%20Examples.html)).

### Data Dump

_No response_

### To Reproduce

Install numpy 2.5.0

### Expected behavior

Should work with this dependency

### Screenshots

_No response_

### Additional context

_No response_

### Community Meeting

- [ ] I wish to discuss this bug in the bi-monthly PGM community meeting. Please join the [PGM community meeting](https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745).

vincentkoppen:
Fixes #282 

The `.__args__` on `NDArray` has changed in numpy v2.5. Also in general is it not recommended to use `__args__` directly, instead we should use `get_args` from typing. 

The current PR checks if the version of numpy is below or above 2.5 and uses the logic that is needed based on this. I think it is also possible to recognize the case without the version check, but think this would make things harder to understand. Let me know whether there are good reasons not to have a version check. 

The tests in this PR are successful with numpy 2.4 and it is also successful in the nightly again: https://github.com/PowerGridModel/power-grid-model-ds/actions/runs/28232535881

Unfortunately all the code is quite "magic" in the sense that there a lot of random `get_args` calls and specific indices that we pick to make it work. I don't see a better way of doing it, given the current setup. 

## Changes
- For the pre-2.5 cases I rewrote the logic to use get_args instead of `__args__`
- Added new logic to support the new 2.5 output

## Remarks
- Sonarcloud isn't happy because two of the paths aren't hit in the coverage. This is the case because we currently still have 2.4 in our lock file, so the if statements for numpy 2.5 are not hit as expected. If/When we update, this will be different. 
_I have added a comment to not track coverage for these if-statements._ 
- Sonarcloud isn't happy because of the complexity of the function. I agree that the function is too complex currently, but have ignored it for now. This is a minimal fix to get things working again.
- We might support a bit less now if people had fancy types, similar to `NDArray3`. Since we now explicitly check whether the second passed argument is a `Literal`. We could loosen this and/or add extra options there, for instance an `int`. But personally would be fine with merging as-is and see if we get any feedback on

### DIFF
--- a/src/power_grid_model_ds/_core/model/arrays/base/array.py
+++ b/src/power_grid_model_ds/_core/model/arrays/base/array.py
@@ -119,10 +119,10 @@ class FancyArray(ABC):  # noqa: B024
             elif is_before_numpy_25 and len(type_args) == 2 and get_origin(type_args[1]) is Literal:  # noqa: PLR2004
                 dtypes[name] = (get_args(get_args(type_args[0])[1])[0], get_args(type_args[1])[0])
             # Expected type_args post-2.5 for NDArray: (numpy.int32,)
-            elif len(type_args) == 1:
+            elif len(type_args) == 1: # pragma: no cover
                 dtypes[name] = type_args[0]
             # Expected type_args post-2.5 for NDArray3: (NDArray[numpy.float64], typing.Literal[3])
-            elif len(type_args) == 2:  # noqa: PLR2004
+            elif len(type_args) == 2:  # noqa: PLR2004 # pragma: no cover
                 dtypes[name] = (get_args(type_args[0])[0], get_args(type_args[1])[0])
             else:
                 raise ValueError(f"dtype {type_def} not understood or supported")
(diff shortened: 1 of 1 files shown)


jaapschoutenalliander (approved):
Lets look at how we can set this function up more cleanly in a follow up PR. Good to have a sollution now

PR #285, opened 2026-06-26
chore: cleanup of get_dtype (with numpy 2.5 support)

vincentkoppen:
A small cleanup of the get_dtype function: 
- Directly determine the errors based on `annotations` instead of the `dtypes` dict
- Remove the `dtypes` dict, directly create the `dtype_list`
  - The string length is now a list comprehension. We could also improve the logic, so that "appending" will check it or something. But this is a first simple cleanup

Should make Sonarcloud happy and make the flow a bit simpler to understand.

### DIFF
--- a/src/power_grid_model_ds/_core/model/arrays/base/array.py
+++ b/src/power_grid_model_ds/_core/model/arrays/base/array.py
@@ -119,10 +119,10 @@ class FancyArray(ABC):  # noqa: B024
             elif is_before_numpy_25 and len(type_args) == 2 and get_origin(type_args[1]) is Literal:  # noqa: PLR2004
                 dtypes[name] = (get_args(get_args(type_args[0])[1])[0], get_args(type_args[1])[0])
             # Expected type_args post-2.5 for NDArray: (numpy.int32,)
-            elif len(type_args) == 1:
+            elif len(type_args) == 1: # pragma: no cover
                 dtypes[name] = type_args[0]
             # Expected type_args post-2.5 for NDArray3: (NDArray[numpy.float64], typing.Literal[3])
-            elif len(type_args) == 2:  # noqa: PLR2004
+            elif len(type_args) == 2:  # noqa: PLR2004 # pragma: no cover
                 dtypes[name] = (get_args(type_args[0])[0], get_args(type_args[1])[0])
             else:
                 raise ValueError(f"dtype {type_def} not understood or supported")
--- a/src/power_grid_model_ds/_core/model/arrays/base/array.py
+++ b/src/power_grid_model_ds/_core/model/arrays/base/array.py
@@ -100,7 +100,7 @@ class FancyArray(ABC):  # noqa: B024
 
     @classmethod
     @lru_cache
-    def get_dtype(cls): # noqa: python:S3776
+    def get_dtype(cls):  # noqa: python:S3776
         annotations = get_public_annotations(cls)
         str_lengths = combine_attribute_from_parent_classes(cls, "_str_lengths", dict)
         dtypes = {}
@@ -119,7 +119,7 @@ class FancyArray(ABC):  # noqa: B024
             elif is_before_numpy_25 and len(type_args) == 2 and get_origin(type_args[1]) is Literal:  # noqa: PLR2004
                 dtypes[name] = (get_args(get_args(type_args[0])[1])[0], get_args(type_args[1])[0])
             # Expected type_args post-2.5 for NDArray: (numpy.int32,)
-            elif len(type_args) == 1: # pragma: no cover
+            elif len(type_args) == 1:  # pragma: no cover
                 dtypes[name] = type_args[0]
             # Expected type_args post-2.5 for NDArray3: (NDArray[numpy.float64], typing.Literal[3])
             elif len(type_args) == 2:  # noqa: PLR2004 # pragma: no cover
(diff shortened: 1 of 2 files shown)


src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -37,6 +37,44 @@
...
+def _parse_annotation_post_25(name: str, type_def: Any, type_args: tuple, str_lengths: dict[str, int]) -> tuple:
+    """Parse an NDArray annotation into a numpy dtype tuple for NumPy >= 2.5."""
+    # Expected type_args for NDArray: (numpy.int32,)
+    if len(type_args) == 1:
+        dtype = type_args[0]
+        return (name, _resolve_str_dtype(name, dtype, str_lengths))
+    # Expected type_args for NDArray3: (NDArray[numpy.float64], typing.Literal[3])
+    if len(type_args) == 2:  # noqa: PLR2004
+        dtype = get_args(type_args[0])[0]
+        return (name, _resolve_str_dtype(name, dtype, str_lengths), get_args(type_args[1])[0])
```
vincentkoppen, line 74: nitpick: personally I would prefer to apply `_resolve_str_dtype` after the parsing. 
So: parse all dtypes -> apply the str filter over all -> return

The str parsing isn't specific to the parsing per numpy version. Now this has to be applied 4 times and is possiby easily forgotten if a new parsing is added.

src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -37,6 +37,44 @@
...
+        try:
+            dtype = get_args(get_args(type_args[0])[1])[0]
+            size = get_args(type_args[1])[0]
+        except IndexError as error:
+            raise ValueError(f"dtype {type_def} not understood or supported") from error
+        return (name, _resolve_str_dtype(name, dtype, str_lengths), size)
+    raise ValueError(f"dtype {type_def} not understood or supported")
+
+
+def _parse_annotation_post_25(name: str, type_def: Any, type_args: tuple, str_lengths: dict[str, int]) -> tuple:
```
vincentkoppen, line 65: nitpick: sonarcloud isn't happy with the return type. I think we could do something like `tuple[str, np.generic | str,...]` or more explicit: `tuple[str, np.generic | str] |  tuple[str, np.generic | str, int]`

The `| str` is only needed if we keep the string conversion. Also it might need `type[np.generic]` instead of `np.generic`, not sure.

src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -37,6 +37,44 @@
...
 
 
+def _resolve_str_dtype(name: str, dtype: Any, str_lengths: dict[str, int]) -> Any:
+    """Replace np.str_ with a fixed-length unicode dtype, leaving other dtypes untouched."""
+    if dtype is np.str_:
+        return np.dtype(f"U{str_lengths.get(name, _DEFAULT_STR_LENGTH)}")
+    return dtype
+
+
+def _parse_annotation_pre_25(name: str, type_def: Any, type_args: tuple, str_lengths: dict[str, int]) -> tuple:
```
vincentkoppen, line 47: Same for this fopr the return type.

src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -37,6 +37,44 @@
 Self = TypeVar("Self", bound="FancyArray")
 
 
+def _resolve_str_dtype(name: str, dtype: Any, str_lengths: dict[str, int]) -> Any:
```
vincentkoppen, line 40: Typing can be improved here as well I think, if `np.generic` or `type[np.generic]` is correct in the previous comments, then it can also be used here.

src/power_grid_model_ds/_core/model/arrays/base/array.py
```diff
@@ -102,45 +140,29 @@ def get_id_columns(cls) -> set[str]:
...
+        parse_annotation = (
+            _parse_annotation_pre_25
+            if version.parse(np.__version__) < version.parse("2.5.0")
+            else _parse_annotation_post_25
+        )
+
+        dtype_list = [
+            parse_annotation(name, type_def, get_args(type_def), str_lengths) for name, type_def in annotations.items()
+        ]
+
```
vincentkoppen, line 165: - nitpick: Prefer to use keyword args
- The fact that we have both type_def and type_args feels a strange. But I think this is needed mainly for the tests right? 
Otherwise it's hard to test both python versions with only passing `type_def`? 
Fine to keep it as is, but I would also be fine with:
```
dtype_list = []
for name, dtype in annotations.items(): 
  try: 
    dtype_args = get_args(dtype)
    dtype_list.append(parse_annotations(name, type_args, ...))
exception: 
   raise ValueError(".... {type_def}....")

vincentkoppen (commented):
Looks good to me overall. 
Just some nitpicks, nothing major. Take a look if you agree with any of them, otherwise this is good to. 
(can't approve, since it's my PR)

Thijss (changes requested):
I have merged https://github.com/PowerGridModel/power-grid-model-ds/pull/321 in the meantime. Is this still relevant?

vincentkoppen:
> I have merged #321 in the meantime. Is this still relevant?

This can be closed.

PR #292, opened 2026-07-17
typing improve `power_grid_model_interface.py`

kornerc:

Typing in `power_grid_model_interface.py` was incomplete.
This PR adds type hinying to all methods of `PowerGridModelInterface`.

### Checks

- [ ] Do you wish to discuss this PR in the bi monthly community meeting?
Please join the [PGM community meeting](<https://zoom-lfx.platform.linuxfoundation.org/meeting/91445523235?password=e565df23-7c6d-4d67-94e6-f5774cfd4745>)

### DIFF
--- a/src/power_grid_model_ds/_core/power_grid_model_interface.py
+++ b/src/power_grid_model_ds/_core/power_grid_model_interface.py
@@ -9,6 +9,7 @@ import warnings
 import numpy as np
 from numpy.typing import NDArray
 from power_grid_model import CalculationMethod, ComponentType, PowerGridModel, initialize_array
+from power_grid_model.data_types import Dataset, SingleDataset
 
 from power_grid_model_ds._core.model.grids.base import Grid
 
@@ -29,18 +30,18 @@ class PowerGridModelInterface:
     def __init__(
         self,
         grid: Grid | None = None,
-        input_data: dict[str, NDArray] | None = None,
+        input_data: SingleDataset | None = None,
         system_frequency: float = 50.0,
     ):
         self.grid = grid or Grid.empty()
         self.system_frequency = system_frequency
 
-        self._input_data: dict[str, NDArray] = input_data or {}
-        self.output_data: dict[str, NDArray] = {}
+        self._input_data: SingleDataset = input_data or {}
+        self.output_data: Dataset = {}
         self.model: PowerGridModel | None = None
 
     @property
-    def input_data(self) -> dict[str, NDArray]:
+    def input_data(self) -> SingleDataset:
         """Get the input data for the PowerGridModel."""
         warnings.warn(
             "Input data has been made private and will be removed as public properety in a future version. "
@@ -50,7 +51,7 @@ class PowerGridModelInterface:
         )
         return self._input_data
 
-    def create_input_from_grid(self) -> dict[str, NDArray]:
+    def create_input_from_grid(self) -> SingleDataset:
         """
         Create input for the PowerGridModel
         """
@@ -84,9 +85,9 @@ class PowerGridModelInterface:
     def calculate_power_flow(
         self,
         calculation_method: CalculationMethod = CalculationMethod.newton_raphson,
-        update_data: dict[str, NDArray] | None = None,
+        update_data: Dataset | None = None,
         **kwargs,
-    ) -> dict[str, NDArray]:
+    ) -> Dataset:
         """Initialize the PowerGridModel and calculate power flow over input data.
 
         If input data is not available, self.create_input_from_grid() will be called to create it.
@@ -108,7 +109,7 @@ class PowerGridModelInterface:
         pgm_array[fields] = internal_array.data[fields]
         return pgm_array
 
-    def update_model(self, update_data: dict[str, NDArray]) -> None:
+    def update_model(self, update_data: Dataset) -> None:
         """
         Updates the power-grid-model using update_data, this allows for batch calculations
 
--- a/src/power_grid_model_ds/_core/power_grid_model_interface.py
+++ b/src/power_grid_model_ds/_core/power_grid_model_interface.py
@@ -7,9 +7,8 @@
 import warnings
 
 import numpy as np
-from numpy.typing import NDArray
 from power_grid_model import CalculationMethod, ComponentType, PowerGridModel, initialize_array
-from power_grid_model.data_types import Dataset, SingleDataset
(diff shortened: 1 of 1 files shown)


src/power_grid_model_ds/_core/power_grid_model_interface.py
```diff
@@ -29,13 +29,13 @@ class PowerGridModelInterface:
     def __init__(
         self,
         grid: Grid | None = None,
-        input_data: dict | None = None,
+        input_data: dict[str, NDArray] | None = None,
```
vincentkoppen: `SingleDataSet` (see main comment of the review)

src/power_grid_model_ds/_core/power_grid_model_interface.py
```diff
@@ -29,13 +29,13 @@ class PowerGridModelInterface:
...
         system_frequency: float = 50.0,
     ):
         self.grid = grid or Grid.empty()
         self.system_frequency = system_frequency
 
-        self._input_data = input_data or {}
+        self._input_data: dict[str, NDArray] = input_data or {}
```
vincentkoppen: `SingleDataSet`

src/power_grid_model_ds/_core/power_grid_model_interface.py
```diff
@@ -50,7 +50,7 @@ def input_data(self) -> dict[str, NDArray]:
         )
         return self._input_data
 
-    def create_input_from_grid(self):
+    def create_input_from_grid(self) -> dict[str, NDArray]:
```
vincentkoppen: `SingleDataSet`

src/power_grid_model_ds/_core/power_grid_model_interface.py
```diff
@@ -84,9 +84,9 @@ def create_grid_from_input_data(self, check_ids: bool = True) -> Grid:
     def calculate_power_flow(
         self,
         calculation_method: CalculationMethod = CalculationMethod.newton_raphson,
-        update_data: dict | None = None,
+        update_data: dict[str, NDArray] | None = None,
```
vincentkoppen: `DataSet`

src/power_grid_model_ds/_core/power_grid_model_interface.py
```diff
@@ -84,9 +84,9 @@ def create_grid_from_input_data(self, check_ids: bool = True) -> Grid:
...
         **kwargs,
-    ):
+    ) -> dict[str, NDArray]:
```
vincentkoppen: `DataSet`

src/power_grid_model_ds/_core/power_grid_model_interface.py
```diff
@@ -100,15 +100,15 @@ def calculate_power_flow(
...
+    def _create_power_grid_array(self, array_name: str) -> NDArray:
         """Create power grid model array"""
         internal_array = getattr(self.grid, array_name)
         pgm_array = initialize_array("input", array_name, internal_array.size)
         fields = self._match_dtypes(pgm_array.dtype, internal_array.dtype)
         pgm_array[fields] = internal_array.data[fields]
         return pgm_array
 
-    def update_model(self, update_data: dict):
+    def update_model(self, update_data: dict[str, NDArray]) -> None:
```
vincentkoppen: `DataSet`

vincentkoppen (commented):
Thanks for this PR, the typing definitely needed improvement! 

`power-grid-model` provides some typing via `power_grid_model.data_types` that is a more complete type hint I think. 
I have discussed it with maintainers of pgm and the comments below should be the correct type hint: 
- input_data -> `SingleDataSet`
- update_data -> `DataSet`
- output_data -> `DataSet`

src/power_grid_model_ds/_core/power_grid_model_interface.py
kornerc: ✅

src/power_grid_model_ds/_core/power_grid_model_interface.py
kornerc: ✅

src/power_grid_model_ds/_core/power_grid_model_interface.py
kornerc: ✅

src/power_grid_model_ds/_core/power_grid_model_interface.py
kornerc: ✅

src/power_grid_model_ds/_core/power_grid_model_interface.py
kornerc: ✅

src/power_grid_model_ds/_core/power_grid_model_interface.py
kornerc: ✅

kornerc:
@vincentkoppen thanks for your clear guidance.
I've applied all your change requests and in addition also `def input_data(self) -> SingleDataset`.

src/power_grid_model_ds/_core/power_grid_model_interface.py
```diff
@@ -100,15 +101,15 @@ def calculate_power_flow(
         )
         return self.output_data
 
-    def _create_power_grid_array(self, array_name: str) -> np.ndarray:
+    def _create_power_grid_array(self, array_name: str) -> NDArray:
```
kornerc: Should the return type remain `NDArray` or should it be changed to `SingleArray | DenseBatchArray` since it calls internally `initialize_array` which returns this union data type?

vincentkoppen: Let's use `SingleArray` as the return type. 

With our use, we always pass an integer to the shape. So with the given overload on `initialize_array` this should always return a `SingleArray`. Also this makes the most sense to me, since we do not have any batch data here. 

-----

Internally it seems that both are just aliases of `np.ndarray`, so there is no real difference:
https://github.com/PowerGridModel/power-grid-model/blob/4db4272ab032f72740bcb4acf880297b1c9b4e45/src/power_grid_model/_core/data_types.py#L23-L26

kornerc: Thanks for the deep dive and your suggestion.
I've applied the change.

vincentkoppen (approved):
Thanks for the PR, really appreciated! 
Approved.

PR #303, opened 2026-07-28
fix: serialize NaN values as JSON null

### LINKED ISSUE #164: [BUG] .json NaN is not officially supported
**Describe the bug**

serializing a Grid to json now converts `float("nan")` values to `NaN`. This is technically not valid json.

Perhaps we can opt for "nan" instead. That is the same solution as PGM has implemented.

Saim-naushad:
Fixes #164

- Serialize floating-point `NaN` values as JSON `null` for standards-compliant output.
- Preserve backward compatibility when deserializing legacy JSON containing bare `NaN`.
- Add regression coverage for file and in-memory JSON serialization.
- Add a legacy JSON fixture containing `NaN`.

### Validation

- `619 passed, 1 skipped`
- Ruff, mypy, and pylint checks passed.

### AI assistance disclosure

I used OpenAI Codex for repository navigation and test planning. I reviewed and understand the submitted changes and can explain the implementation.

- [ ] Do you wish to discuss this PR in the bi-monthly community meeting?


src/power_grid_model_ds/_core/model/grids/serialization/json.py
```diff
@@ -182,3 +183,14 @@ def _is_serializable(value: Any, strict: bool, **kwargs) -> bool:
...
+
+def _replace_nan_with_none(value: Any) -> Any:
+    """Recursively replace NaN values with JSON-compatible null values."""
+    if isinstance(value, float) and math.isnan(value):
+        return None
+    if isinstance(value, dict):
+        return {key: _replace_nan_with_none(item) for key, item in value.items()}
+    if isinstance(value, list):
+        return [_replace_nan_with_none(item) for item in value]
+    return value
```
vincentkoppen, line 192: I think it's possible and cleaner to fix this directly when we create the data instead of a post-processing step. 

Specifically I think we should fix it in `_serialize_array`. 
https://github.com/PowerGridModel/power-grid-model-ds/blob/db1a4c69a486ace1f1348ef4ec9a70e6a4594731/src/power_grid_model_ds/_core/model/grids/serialization/json.py#L146-L147

We could change it so that we: 
- Have the helper `_replace_nan_with_none` as you currently have but without the two `dict/list` if-statements
- Change `_serialize_array` to `return [{name: _replace_nan_with_none(record[name].item()) for ...`

vincentkoppen (commented):
Thanks for the PR 😄  
I have one request with regards to how/where we fix this issue.

src/power_grid_model_ds/_core/model/grids/serialization/json.py
Saim-naushad: Thanks! I’ve moved the conversion into `_serialize_array()` and simplified `_replace_nan_with_none()` to handle scalar values only, as suggested. The full suite still passes with 619 passed and 1 skipped, along with Ruff, mypy, pylint, and formatting checks.

vincentkoppen (approved):
Thanks for the PR, ready to go! :)

PR #311, opened 2026-08-12
Minor fixes: from_grid for custom graph container, create_grid_from_input_data clears exsiting grid.

nitbharambe:
- Allow empty to re-recreate custom graphs
- create_grid_from_input_data clears exsiting grid

### DIFF
--- a/src/power_grid_model_ds/_core/model/graphs/container.py
+++ b/src/power_grid_model_ds/_core/model/graphs/container.py
@@ -135,7 +135,8 @@ class GraphContainer:
         """Build from grid"""
         cls._validate_branches(arrays=grid)
 
-        new_container = cls.empty()
+        graph_model_type = type(grid.graphs.active_graph)
+        new_container = cls.empty(graph_model=graph_model_type)
         for graph_field in new_container.graph_attributes:
             graph: BaseGraphModel = getattr(new_container, graph_field.name)
             new_graph = graph.from_grid(grid, active_only=graph.active_only)
--- a/src/power_grid_model_ds/_core/power_grid_model_interface.py
+++ b/src/power_grid_model_ds/_core/power_grid_model_interface.py
@@ -72,13 +72,19 @@ class PowerGridModelInterface:
 
         Returns a Grid object with the arrays filled with the PowerGridModel input.
         """
+        graph_model_type = type(self.grid.graphs.active_graph)
+        new_grid = self.grid.__class__.empty(graph_model=graph_model_type)
+
         for pgm_name in ComponentType:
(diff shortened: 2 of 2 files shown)


src/power_grid_model_ds/_core/power_grid_model_interface.py
```diff
@@ -73,11 +73,16 @@ def create_grid_from_input_data(self, check_ids: bool = True) -> Grid:
...
         for pgm_name in ComponentType:
-            if pgm_name in self._input_data and hasattr(self.grid, pgm_name):
-                self.grid.append(self._create_pgm_ds_array(pgm_name), check_max_id=False)
+            if pgm_name in self._input_data and hasattr(new_grid, pgm_name):
+                new_grid.append(self._create_pgm_ds_array(pgm_name), check_max_id=False)
         if check_ids:
-            self.grid.check_ids()
+            new_grid.check_ids()
+
+        self.grid = new_grid
```
vincentkoppen, line 96: I agree that this is the intended behaviour of this method. It could cause problems for existing users of the function, but this is not really likely. The only case I could think of is: 

- A user has a grid with an additional array/object on the grid 
- Initializes the interface
- Already adds the extra array/object on the `interface.grid`
- Now calls create_grid_from_input_data
  - This change now clears the Grid

 So I'm fine with changing this, what do you think @Thijss  and @jaapschoutenalliander ?

vincentkoppen: - Might be nice to adjust the docstring to clearly state that this will overwrite and (re-)initialize the grid based on input_data. 
- We could build in a check that crashes/gives a warning if `self.grid` is not empty at the start of this function? Or even a deprecation warning?

vincentkoppen (commented):
Looks good to me, let's see if the others agree.

src/power_grid_model_ds/_core/power_grid_model_interface.py
nitbharambe: Added docstring and warning. Better than silently changing behaviour. Fine with crashing as well if we decide on it.

src/power_grid_model_ds/_core/model/graphs/container.py
```diff
@@ -135,7 +135,8 @@ def from_grid(cls, grid: "Grid") -> "GraphContainer":
         """Build from grid"""
         cls._validate_branches(arrays=grid)
 
-        new_container = cls.empty()
+        graph_model_type = type(grid.graphs.active_graph)
```
Thijss, line 138: Could you elaborate what the new code makes possible? I don't see yet why the change is necessary.
What was going wrong before?

